Issue #32, opened 2015-01-05
Propagate content type for SNS requests


rpk:
NotificationMessagingTemplate convertAndSend accepts a set of headers but later does not propagate the full set in TopicMessageChannel.send.
```java
@Override
public <T> void convertAndSend(String destinationName, T payload, Map<String, Object> headers) throws MessagingException {
	Assert.isInstanceOf(String.class, payload, "Payload must be of type string");
	super.convertAndSend(destinationName, payload, headers);
}
```
[TopicMessageChannel.java#L45-L50](https://github.com/spring-cloud/spring-cloud-aws/blob/master/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/core/TopicMessageChannel.java#L45-L50)
```java
@Override
public boolean send(Message<?> message, long timeout) {
	this.amazonSns.publish(new PublishRequest(this.topicArn,
			message.getPayload().toString(), findNotificationSubject(message)));
	return true;
}
```
For QueueMessagingTemplate to receive json formatted messages sent by the NotificationMessageTemplate they need to have an application/json contentType which will never be propagated as the TopicMessageChannel ignores the contentType header.

Propagating the content-type header from the message into the publish request as a message attribute will allow QueueMessageTemplate to correctly deserialize json encoded types.

QueueMessageTemplate performs a similar process when sending messages:
[QueueMessageChannel.java#L55-L80](https://github.com/spring-cloud/spring-cloud-aws/blob/master/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/core/QueueMessageChannel.java#L55-L80)
```java
Map<String, MessageAttributeValue> messageAttributes = getContentTypeMessageAttributes(message);
if (!messageAttributes.isEmpty()) {
	sendMessageRequest.withMessageAttributes(messageAttributes);
}
```

Also, it might be useful to note in the docs that in order to enable this use case, raw message delivery must be set to True.

> Important
>To use message attributes with Amazon SQS endpoints, you must set the subscription attribute, Raw Message Delivery, to True. For more information about Raw Message Delivery, see Appendix: Large Payload and Raw Message Delivery

---

Until content-type is propagated, we are using the following workaround:
In the consumer, add a custom argument resolver:
```xml
<aws-messaging:annotation-driven-queue-listener wait-time-out="20">
	<aws-messaging:argument-resolvers>
		<bean class="com.bar.FooResolver"/>
[...]

Issue #12, opened 2014-11-12
Support SQS MessageAttributes


jankronquist:
Currently only "contentType" attribute is supported.

PR #37, opened 2015-01-13
Feature/message attributes

alainsahli:
Solves #12 and #32.

### DIFF
--- a/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/core/QueueMessageChannel.java
+++ b/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/core/QueueMessageChannel.java
@@ -26,10 +26,8 @@ import com.amazonaws.services.sqs.model.SendMessageRequest;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageDeliveryException;
 import org.springframework.messaging.MessageHeaders;
-import org.springframework.messaging.MessagingException;
 import org.springframework.messaging.PollableChannel;
 import org.springframework.messaging.support.AbstractMessageChannel;
-import org.springframework.messaging.support.GenericMessage;
 import org.springframework.util.Assert;
 import org.springframework.util.MimeType;
 import org.springframework.util.NumberUtils;
@@ -37,6 +35,8 @@ import org.springframework.util.NumberUtils;
 import java.nio.ByteBuffer;
 import java.util.HashMap;
 import java.util.Map;
+
+import static org.springframework.cloud.aws.messaging.core.QueueMessageUtils.createMessage;
 
 /**
  * @author Agim Emruli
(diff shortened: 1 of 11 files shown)


spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/core/QueueMessageChannel.java
```diff
@@ -41,9 +45,8 @@
  */
 public class QueueMessageChannel extends AbstractMessageChannel implements PollableChannel {
 
-	static final String MESSAGE_RECEIVING_ATTRIBUTE_NAMES = "All";
-	private static final String RECEIPT_HANDLE_MESSAGE_ATTRIBUTE_NAME = "ReceiptHandle";
-	private static final String MESSAGE_ID_MESSAGE_ATTRIBUTE_NAME = "MessageId";
+	static final String ATTRIBUTE_NAMES = "All";
```
aemruli: Why do we have both times this constant ?

spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/core/QueueMessageChannel.java
```diff
@@ -68,20 +71,53 @@ protected boolean sendInternal(Message<?> message, long timeout) {
...
 
-	private Map<String, MessageAttributeValue> getContentTypeMessageAttributes(Message<?> message) {
-		Map<String, MessageAttributeValue> messageAttributes = new HashMap<>(1);
-		Object mimeType = message.getHeaders().get(MessageHeaders.CONTENT_TYPE);
-		if (mimeType != null) {
-			if (mimeType instanceof MimeType) {
-				messageAttributes.put(MessageHeaders.CONTENT_TYPE, new MessageAttributeValue().withDataType("String").withStringValue(mimeType.toString()));
-			} else if (mimeType instanceof String) {
-				messageAttributes.put(MessageHeaders.CONTENT_TYPE, new MessageAttributeValue().withDataType("String").withStringValue((String) mimeType));
+	private Map<String, MessageAttributeValue> getMessageAttributes(Message<?> message) {
```
aemruli: make this method static (as we do not use any instance fields) ?

spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/core/QueueMessageChannel.java
alainsahli: Because on a SQS message there are attributes and message names (two different things). It's only a "coincidence" that the constant value is the same. But the name of the constants is confusing, I'm going to change it.

spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/core/QueueMessageChannel.java
alainsahli: Actually this method uses the logger which is an instance field. So I'll leave it as is.

PR #39, opened 2015-01-14
With this commit if instance meta-data returned custom region will try t...

gschukin:
...o map it to region meta-data not Regions enum


aemruli:
@gschukin Thanks

PR #44, opened 2015-01-16
Cache OnAwsCloudEnvironmentCondition first result to avoid multiple http requests for every check

gschukin:
Prior to this commit every time @ConditionalOnAwsCloundEnvironment runs its check http request was performed.

With this commit first check result is cached and then returned for every next check.


aemruli:
Thanks @gschukin

PR #57, opened 2015-02-17
Add support to SimpleStorageResourceLoader for versioned S3 objects

robinhowlett:
Versioning in S3 is a means of keeping multiple variants of an object
in the same bucket.

This commit adds support for loading resources of specific versions of
objects within a version-enabled bucket
e.g. `"s3://bucket/object^versionId"`

- Added version delimiter (`^`) to separate object names from version Ids
- Changed Amazon S3 clients use of `getObject()` and `getObjectMetadata()` to
  use `GetObjectRequest` and `GetObjectMetadataRequest` objects rather than
  just `"bucket"` and `"object"`
- Updated unit tests to match on use of `GetObjectRequest` and
  `GetObjectMetadataRequest` types rather than specific requests of `"bucket"`
  and `"object"`


aemruli:
@robinhowlett Robin thanks for the PR. Did you sign the CLA ?

robinhowlett:
@aemruli Yes, just signed, conf. number 109820150225051513

aemruli:
Thanks @robinhowlett

Issue #59, opened 2015-03-08
SimpleMessageListenerContainer eats up exceptions without logging them


rahulpilani:
On line 194 of SimpleMessageListenerContainer, after the catching the exception, we should at least log it, otherwise, anyone looking at the logs will have no clue why a given message is being re-processed or thrown into the DLQ.

rahulpilani:
Opened up a PR for this here: 
https://github.com/spring-cloud/spring-cloud-aws/pull/60

aemruli:
Merged pull request

PR #60, opened 2015-03-08
Logging the error after catching it

rahulpilani:
Otherwise logs will have no trace of the exception.

Fix for:
https://github.com/spring-cloud/spring-cloud-aws/issues/59


rahulpilani:
I have signed and agree to the terms of the SpringSource Individual
Contributor License Agreement.

PR #76, opened 2015-05-29
Fix minor documentation error

Sintendo:


aemruli:
@Sintendo thanks for spotting

PR #90, opened 2015-08-11
Allow for the use of temporary credentials

jkschneider:
This is especially useful when running the integration tests.  There is no need to establish a new IAM user.  Temporary credentials can be taken from any running instance with:

`curl -s http://169.254.169.254/latest/meta-data/iam/security-credentials`

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -71,8 +71,9 @@ configuration files that configure the necessary parameters to build the
 environment.
 
 Please create a new file named access.properties. This file must contain
-three properties named accessKey,secretKey and rdsPassword. These two
-properties accessKey and secretKey are account/user specific and should
+three properties named accessKey,secretKey and rdsPassword.  If you are
+using temporary credentials, it must also contain sessionToken.  The
+accessKey and secretKey properties are account/user specific and should
 never be shared to anyone. To retrieve these settings you have to open
 your account inside the AWS console and retrieve them through the
 https://portal.aws.amazon.com/gp/aws/securityCredentials[Security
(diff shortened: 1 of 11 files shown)


adriancole:
LGTM

aemruli:
I had errors while running the integration tests 

Caused by: com.amazonaws.AmazonServiceException: The security token included in the request is invalid (Service: AmazonCloudFormation; Status Code: 403;

Will check them tomorrow, anyway thanks @jkschneider for the PR.

jkschneider:
@aemruli it may be that temporary credentials can never run Cloud Formation scripts, not sure.

jkschneider:
BTW, there is a potentially larger use of this, which is to make the whole SDK credentials configuration block optional.  The assumption is this: if the service is running on an AWS instance, it can query `http://169.254.169.254/latest/meta-data/iam/security-credentials` itself, and populate the SDK credentials automatically.

Thoughts?

aemruli:
Hi @jkschneider, I had a deeper review into the PR and also your last comment. I am currently asking myself if the implementation of this PR is feasible for users. As far as i understand, temporary session credentials are valid only for a short period (1 to 36 hours). Therefore a configured session key would become invalid at the latest 36 hours after application bootstrap. 

If we want to allow users to use temporary credentials, then I think we must use a com.amazonaws.auth.STSSessionCredentialsProvider which periodically refreshes the temporary credentials based on configured long-lived credentials. 

Did I miss something here ?

As for your last comment, I think the instance-profile credentials provider will do exactly that (which is already available).

ryangardner:
I went another route to get support for temporary credentials in #170  - spring's `EnvironmentVariableCredentialsProvider` already supports temporary credentials, so if you use the DefaultAWSCredentialsProviderChain instead of the one that spring boot is creating on its own you get that for free. 

(The reason I need to use temporary credentials is for local development - those are the only credentials that developers get from the credential management system we use here: https://github.com/Cox-Automotive/ALKS )

maciejwalkowiak:
Fixed by #170.

PR #96, opened 2015-08-28
Add redirect support to AmazonS3 operations

gregturn:
Doing s3 operations in buckets found in different regions causes AmazonS3 SDK to throw an AmazonS3Exception. This patch wraps AmazonS3 in a proxy that traps for 301 exceptions, parses the error response XML for the proper EndPoint, updates amazonS3, and tries again.

### DIFF
--- a/spring-cloud-aws-core/src/main/java/org/springframework/cloud/aws/core/io/s3/AmazonS3ProxyFactory.java
+++ b/spring-cloud-aws-core/src/main/java/org/springframework/cloud/aws/core/io/s3/AmazonS3ProxyFactory.java
@@ -1,0 +1,145 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


gregturn:
@dsyer @marcingrzejszczak could you please review my PR?

spring-cloud-aws-core/src/main/java/org/springframework/cloud/aws/core/io/s3/AmazonS3ProxyFactory.java
```diff
@@ -0,0 +1,145 @@
...
+import com.amazonaws.services.s3.model.AmazonS3Exception;
+import com.amazonaws.util.StringInputStream;
+
+/**
+ * Proxy to wrap an {@link AmazonS3} handler and handle redirects wrapped inside {@link AmazonS3Exception}.
+ *
+ * @author Greg Turnquist
+ * @since 1.1
+ */
+public class AmazonS3ProxyFactory {
```
artembilan: I would like to see something like `AmazonS3ProxyFactoryBean` to let any S3 client to get a gain from the redirect workaround.

OTOH shouldn't we raise an issue against AWS to utilize the feature out-of-the-box?
Thanks

gregturn:
@artembilan After digging in, I can't really see a solution involving a ProxyFactoryBean (yet). Is it possible to go ahead and use this? Perhaps such a factory bean would be more suitable in a Spring Boot autoconfigure policy?

mminella:
@ alainsahli or @ bountin What can we do to get this merged in?

spencergibb:
@mminella @gregturn if we don't hear from them soon, I can do it. Especially if you want this in the milestone we're getting ready to release.

gregturn:
@spencergibb Yup. I have basically put this project on the shelf as unusable regarding this feature.

Issue #114, opened 2015-10-30
Support multiple CloudFormation stacks


sfussenegger:
(this is related to #113)

The current implementation of `ResourceIdResolver` already has a TODO comment for this one: "discuss whether to support more than one stack resource registry and how to deal with ordering/name conflicts"

As suggested in #113, this could be solved by qualifying resources with the stack name, allowing for unqualified references of unique names only.

[Logical ID must be alphanumeric (A-Za-z0-9)](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resources-section-structure.html) and [a stack name can contain only alphanumeric characters (case sensitive) and hyphens](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-using-console-create-stack-parameters.html). Therefore, there are several characters available, e.g. `stack.resource` or `stack/resource`

Issue #113, opened 2015-10-30
Support nested CloudFormation stacks in StackResourceRegistryFactoryBean


sfussenegger:
We use nested stacks quite a bit to make templates reusable. It would be great to have `ResourceIdResolver` support this. 

I've just made a prototype work extending `StackResourceRegistryFactoryBean` by joining logical resource ids, e.g. `NestedStack` and `NestedStack.NestedResource` or `NestedStack.AnotherNestedStack.AnotherResource`.

PS: Qualifying resources similarly with the stack name this could also be the solution for the `TODO` in `StackResourceRegistryDetectingResourceIdResolver`: "discuss whether to support more than one stack resource registry and how to deal with ordering/name conflicts". So the name of a resource would be `{stackName}.{logicalResourceId}` or simply `{logicalResourceId}` if the name is unique.

PPS: Honestly, it was quite annoying to work on this stuff as `StackResourceRegistryFactoryBean` doesn't really allow for customization: a lot of private methods, no access to private fields, package-private classes etc. I had to put classes in `org.springframework.*` packages just circumvent package-privacy and copy quite some private code. For instance, `StackResourceRegistryFactoryBean.convertToStackResourceMappings(..)` is `private static` while it could easily be `protected` for customization. Another example are both `StackNameProvider` implementations that are package-private, making it unnecessarily complicated to create a completely custom `ListableStackResourceFactory` implementation.

aemruli:
This should be solved already now with #124 , otherwise please re-open.

PR #124, opened 2016-01-25
#113 Support nested CloudFormation stacks in StackResourceRegistryFactoryBean

sfussenegger:
Support nested CloudFormation stacks by creating a path of logical resource ids ('.' isn't allowed in CloudFormation stack or resource names, see #114). 

As an example, a CFM stack could contain two nested stacks named "RdsStack" and "CacheStack" and resource lookup would work using an id like "RdsStack.RdsInstance" and "CacheStack.CacheInstance"

This PR also adds a feature to access nested stack resources by their unqualified name as long as it is unique ("RdsInstance" or "CacheIsntance" in the above example). This could also be used to support multiple non-nested stacks (again see #114)


aemruli:
@sfussenegger thanks!

Issue #125, opened 2016-01-25
AutoDetectingStackNameProvider should use instance tags to support auto scaling groups


sfussenegger:
the current implementation describes stack resources using the EC2 instance ID as a physical resource ID. This does not work for instances that are part of an auto scaling group. As a fallback, the implementation should try to read the `aws:cloudformation:stack-name` tag.

PR #126, opened 2016-01-25
#125 AutoDetectingStackNameProvider should use instance tags to support auto scaling groups

sfussenegger:
the main change is in `AutoDetectingStackNameProvider` and rather simple at that. If an `EC2Client` is available, it is used to describe the instance tags. Missing permissions to do so would require in an exception. This could be handled better but it is the same as for missing permissions to describe stack resources.

It wasn't easy to get the `AmazonEC2` client injected which is what caused all the other changes:

- `ContextStackConfiguration` uses autowired fields instead of methods
- `ContextStackAutoConfiguration` makes better reuse of the code from `ContextStackConfiguration`, only changing the way a static stack name is configured
- `StackNameProvider` and its implementation are now public, making it easier to customize
- `StackConfigurationBeanDefinitionParser` added attribute `amazon-ec2`

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextStackAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextStackAutoConfiguration.java
@@ -16,7 +16,6 @@
 
 package org.springframework.cloud.aws.autoconfigure.context;
 
-import com.amazonaws.services.cloudformation.AmazonCloudFormation;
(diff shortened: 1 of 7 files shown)


sfussenegger:
looks like the TravisCI build failed because of a race condition. `BufferingCloudWatchMetricSenderTest` has nothing to do with the PR, even though the failed test interestingly is named  `send_withMetricData_sendsDataToCloudFormation()` even though it should really be named `send_withMetricData_sendsDataToCloudWatch()`. The test sleeps for 15 seconds and checks for an async event.

aemruli:
Thanks @sfussenegger, allthough I had to do some change in order to make the integration tests run again.

PR #163, opened 2016-07-14
Conditional on aws cloud environment

mbenson:
As discussed on gitter, these changes preclude any of the spring-cloud-aws autoconfiguration from running when included in a non-AWS environment. I have signed the CLA.

_Edit:_ I see that the new CLA allows the system to know that automatically. Nice!

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextCredentialsAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextCredentialsAutoConfiguration.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.aws.autoconfigure.context;
 
 import com.amazonaws.auth.profile.ProfilesConfigFile;
 import org.springframework.beans.factory.support.BeanDefinitionRegistry;
+import org.springframework.cloud.aws.context.annotation.ConditionalOnAwsCloudEnvironment;
 import org.springframework.cloud.aws.context.config.annotation.ContextDefaultConfigurationRegistrar;
 import org.springframework.context.EnvironmentAware;
 import org.springframework.context.annotation.Configuration;
@@ -32,6 +33,7 @@ import static org.springframework.cloud.aws.context.config.support.ContextConfig
  * @author Agim Emruli
  */
 @Configuration
+@ConditionalOnAwsCloudEnvironment
 @Import({ContextDefaultConfigurationRegistrar.class, ContextCredentialsAutoConfiguration.Registrar.class})
 public class ContextCredentialsAutoConfiguration {
 
--- a/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextRegionProviderAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextRegionProviderAutoConfiguration.java
@@ -17,6 +17,7 @@
 package org.springframework.cloud.aws.autoconfigure.context;
 
 import org.springframework.beans.factory.support.BeanDefinitionRegistry;
+import org.springframework.cloud.aws.context.annotation.ConditionalOnAwsCloudEnvironment;
 import org.springframework.context.EnvironmentAware;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
@@ -30,6 +31,7 @@ import static org.springframework.cloud.aws.context.config.support.ContextConfig
  * @author Agim Emruli
  */
 @Configuration
+@ConditionalOnAwsCloudEnvironment
 @Import(ContextRegionProviderAutoConfiguration.Registrar.class)
 public class ContextRegionProviderAutoConfiguration {
 
--- a/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextResourceLoaderAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextResourceLoaderAutoConfiguration.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.aws.autoconfigure.context;
 
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.support.BeanDefinitionBuilder;
+import org.springframework.cloud.aws.context.annotation.ConditionalOnAwsCloudEnvironment;
 import org.springframework.cloud.aws.context.config.annotation.ContextResourceLoaderConfiguration;
 import org.springframework.context.EnvironmentAware;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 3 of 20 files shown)


spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextCredentialsAutoConfiguration.java
```diff
@@ -32,6 +33,7 @@
  * @author Agim Emruli
  */
 @Configuration
+@ConditionalOnAwsCloudEnvironment
```
artembilan: I don't agree with this one. 
Why should I need EC2 if I just would like just to have access to AWS ?

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextRegionProviderAutoConfiguration.java
```diff
@@ -30,6 +31,7 @@
  * @author Agim Emruli
  */
 @Configuration
+@ConditionalOnAwsCloudEnvironment
```
artembilan: Not sure about this one, too.
Actually I have an integration test based on CloudFormation and stack and I do testing without EC2 and even its mocking.

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextResourceLoaderAutoConfiguration.java
```diff
@@ -29,6 +30,7 @@
  * @author Agim Emruli
  */
 @Configuration
+@ConditionalOnAwsCloudEnvironment
```
artembilan: The same concern is here.

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextStackAutoConfiguration.java
```diff
@@ -33,6 +33,7 @@
  * @author Agim Emruli
  */
 @Configuration
+@ConditionalOnAwsCloudEnvironment
```
artembilan: H-m...
I would like to see that discussion on Gitter, because allmost all these changes are breaking for me.
Looks like I fully can't get a gain from AWS auto-configuration in my Boot apps if I don't use EC2. :cry:

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextCredentialsAutoConfiguration.java
mbenson: I am quite new to AWS and have taken a "shotgun" approach here. After a dive through the code I agree this one does not look as though it will explode outside the AWS environment. Removing the condition here does not seem to make my application fail when run locally.

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextStackAutoConfiguration.java
mbenson: Okay, maybe we can catch up on gitter tomorrow and perhaps you can explain the fundamental things that I'm missing, as far as you'd like. :) I'll narrow down precisely which autoconfiguration class was stopping me from even starting my application. Thanks!

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextResourceLoaderAutoConfiguration.java
mbenson: This one definitely stops me from starting my app in my local, non-AWS environment, ultimately because region auto-detection is being used (by default). I'm around today on gitter if you want to discuss.

mbenson: I have experimented here with a new `@ConditionalOnAwsRegionAvailable` implementation that tries the available `RegionProvider`; if a static region is set the condition can pass outside EC2.

artembilan: Right, that was my intention.
If we use properties like:
```
cloud.aws.region.static
cloud.aws.stack.name
```
we shouldn't require EC2 environment.

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextStackAutoConfiguration.java
artembilan: No, I don't think that Gitter would help somehow.
It's even much better that we have a code scope in this Pr already, that may simplify a discussion.

So, this change is definitely wrong.
See source code of this  `ContextStackAutoConfiguration` class, especially part like:
```
@ConditionalOnProperty(prefix = "cloud.aws", name = "stack.auto", havingValue = "true", matchIfMissing = true)
@ConditionalOnAwsCloudEnvironment
@ConditionalOnMissingBean(StackResourceRegistryFactoryBean.class)
public static class StackAutoDetectConfiguration extends ContextStackConfiguration {
```
So, we have already `@ConditionalOnAwsCloudEnvironment` there and it is enough IMO.

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextRegionProviderAutoConfiguration.java
artembilan: This one is wrong as well.
See its source code:
```
registerRegionProvider(registry, this.environment.getProperty("cloud.aws.region.auto", Boolean.class, true) &&
	!(this.environment.containsProperty("cloud.aws.region.static")),
	this.environment.getProperty("cloud.aws.region.static"));
```
So, we register `RegionProvider` if it is `auto` - `Ec2MetadataRegionProvider`, or if we explicitly provide `cloud.aws.region.static`.
Otherwise there will no any `RegionProvider` independently of the environment.

artembilan:
OK. I'm not well with many AWS services, like `CloudWatch` or `Elasti Cache`, but looks like we don't need this change at all.
According to our discussion that feels like you have just missed two properties mentioned above:
```
cloud.aws.region.static
cloud.aws.stack.name
```

Everything rest is fine. And it isn't obvious by their code that EC2 is a requirement.

I'm sure @aemruli and @alainsahli can provide some feedback on the matter after a signal with this PR and their wisdom in those classes.

Anyway thank you for contribution and good discussion.
There are much more which require some fix. :smile: 

Don't hesitate to go ahead with other PRs!

mbenson:
As this is nearly 4 years old we can chalk this up to lack of experience on
my part with AWS at the time.

maciejwalkowiak:
Understood 😉

PR #170, opened 2016-09-08
Allow Spring Boot users to use the DefaultAWSCredentialsProviderChain

ryangardner:
Amazon [recommends using the DefaultAWSCredentialsProviderChain](http://docs.aws.amazon.com/java-sdk/latest/developer-guide/credentials.html) instead of creating a custom chain. 

The custom chain that is created has some limitations (notably for me, the lack of support for using session tokens) - rather than invest time in mapping custom spring properties, I would rather just use the DefaultAWSCredentialsProviderChain which allows me to specify environment variables for my access credentials when running locally and will use an InstanceProfile when running on my EC2 instance.

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-aws.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-aws.adoc
@@ -285,7 +285,14 @@ Spring Cloud AWS provides the following properties to configure the credentials 
 |cloud.aws.credentials.instanceProfile
 |true
 |Configures an instance profile credentials provider with no further configuration
+
+|cloud.aws.credentials.useDefaultAwsCredentialsChain
+|true
+|Use the DefaultAWSCredentials Chain instead of configuring a custom credentials chain
 |===
+
+
+
 
 ===== Configuring region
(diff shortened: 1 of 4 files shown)


guidograzioli:
Running the four lines you changed in ContextConfigurationUtils like 

```
   @Override
    public void postProcessBeanDefinitionRegistry(BeanDefinitionRegistry registry)
            throws BeansException {
  	  BeanDefinitionBuilder builder = BeanDefinitionBuilder.rootBeanDefinition(DefaultAWSCredentialsProviderChain.class);
  	  registry.registerBeanDefinition(CredentialsProviderFactoryBean.CREDENTIALS_PROVIDER_BEAN_NAME, builder.getBeanDefinition());
  	  AmazonWebserviceClientConfigurationUtils.replaceDefaultCredentialsProvider(registry, CredentialsProviderFactoryBean.CREDENTIALS_PROVIDER_BEAN_NAME);
    }
```

was the only way for me to run spring-boot in an AWS lambda and call into other amazon apis with the role associated to the lambda. (your pull request #90 would also do the job eventually).

PR #185, opened 2016-11-15
Don't autoconfigure SES if already configured

candrews:
If a MailSender is already configured, don't configure SES.

This change allows the user to specify non-SES mail configuration, or override the SES mail configuration, without having to disable MailSenderAutoConfiguration.

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/mail/MailSenderAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/mail/MailSenderAutoConfiguration.java
@@ -21,6 +21,7 @@ import com.amazonaws.services.simpleemail.AmazonSimpleEmailService;
 import com.amazonaws.services.simpleemail.AmazonSimpleEmailServiceClient;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingClass;
 import org.springframework.cloud.aws.autoconfigure.context.ContextCredentialsAutoConfiguration;
 import org.springframework.cloud.aws.context.annotation.ConditionalOnMissingAmazonClient;
@@ -40,6 +41,7 @@ import javax.mail.Session;
  */
 @Configuration
 @ConditionalOnClass({MailSender.class, AmazonSimpleEmailService.class})
+@ConditionalOnMissingBean(MailSender.class)
 @Import(ContextCredentialsAutoConfiguration.class)
 public class MailSenderAutoConfiguration {
 


spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/mail/MailSenderAutoConfiguration.java
```diff
@@ -40,6 +41,7 @@
  */
 @Configuration
 @ConditionalOnClass({MailSender.class, AmazonSimpleEmailService.class})
+@ConditionalOnMissingBean(MailSender.class)
```
artembilan: Hello, @candrews !

Thank you for the PR, but to be honest I don't like this approach.
Although I see that it is typical in Spring Boot. For example similar `org.springframework.boot.autoconfigure.mail.MailSenderAutoConfiguration` has the same `@ConditionalOnMissingBean(MailSender.class)`.

So, right now, when I have both these configuration in class path, the regular `JavaMailSenderImpl` `@Bean` is overridden by this SES `MailSender`.

Please, describe how would you deal if we need both `MailSender`s? For Google and AWS, for example? Or that does not make sense and we really should use only one?

From here, otherwise, we might consider something like `@AutoConfigureBefore(org.springframework.boot.autoconfigure.mail.MailSenderAutoConfiguration.class)` to let end-user solution win and override the Boot solution at the same time, if we win here.

candrews: I think these autoconfigurers are intended to provide sensible defaults that users can override. Assuming that's true, it's surprising when a user configures a `MailSender` bean and then `org.springframework.cloud.aws.autoconfigure.mail.MailSenderAutoConfiguration` overrides it with no (obvious) way to stop it from doing so.

I also think the purpose of the autoconfigurers is to serve the more common cases and not solve for everything. In your example, having mail senders configured for both Google and AWS at the same time is (I think, at least) highly unusual - an application almost always has exactly 1 mail sender. For those users who want multiple, they can certainly do so, but not using the autoconfigurers.

candrews: `org.springframework.boot.autoconfigure.mail.MailSenderAutoConfiguration.class` will only do something if the configuration properties `spring.mail.host` or `spring.mail.jndi-name` are set, see https://github.com/spring-projects/spring-boot/blob/v1.4.2.RELEASE/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mail/MailSenderAutoConfiguration.java#L105 I think if a user set those properties, then they want to configure mail and not rely on SES.

So actually, I think doing the opposite of what you're suggesting the is the way to go, add:
`@AutoConfigureAfter(org.springframework.boot.autoconfigure.mail.MailSenderAutoConfiguration.class)`

artembilan: Oh! Didn't notice that. 
Right, that would be better.
`org.springframework.boot.autoconfigure.mail.MailSenderAutoConfiguration` has the same `@ConditionalOnMissingBean(MailSender.class)`, so making it `after` would collide with SES independently of the `spring.mail.*` properties. 
With what you suggest we can still have SES if we don't use `spring.mail.*` properties, and can override SES, if we would like to use `spring.mail.*` and still rely just on the default Spring Boot `MailSender`.
In case of both we don't have choice unless configure everything manually. Because `@ConditionalOnMissingBean(MailSender.class)` kills both auto-configuration, when we have even only just one custom `MailSender`.

Am I correct?

candrews: That matches my thinking :+1:

candrews:
Any news? I'd love to see this issue get resolved.

artembilan:
@candrews ,

Sorry for delay.
Looks like this project is a bit out of support...

Anyway let's try to make our best!

I think we can't apply this change to the maintenance version `1.1.x` because that would be like a breaking change.
Let's move it to a new PR against `master` - `1.2.0` and, please, add `@AutoConfigureAfter(org.springframework.boot.autoconfigure.mail.MailSenderAutoConfiguration.class)` as we discussed.

Mention me there and I'll merge that PR.

Thank you for contribution!

candrews:
Here's the pull request against master: https://github.com/spring-cloud/spring-cloud-aws/pull/203

PR #213, opened 2017-04-14
fix for default task executor having unbounded max pool size

demboos:
The default task executor created in SimpleMessageListenerContainer has in all circumstances undefined max pool size. While this might not be extremely bad, it's a bit unnecessary and misleading. 

The reason is that inside the SimpleMessageListenerContainer.initialize method, the task executor is created before the call to super.initialize(). The createDefaultTaskExecutor() method uses the number of registered queues (getRegisteredQueues().size()) as the base for defining the max pool size, but the list of registered queues is created only when super.initialize() is called. So the order is wrong.

Added to that, the computed max pool size is wrong (spinningThreads *  maxNumberOfMessagePerBatch), because the SimpleMessageListenerContainer uses an extra thread per registered queue for fetching the messages. The proper number should be 

spinningThreads * (maxNumberOfMessagePerBatch + 1)

If only the first fix is applied (moving super.initialize() to the top of SimpleMessageListenerContainer.initialize), the threadpool will reject tasks, since it's queue size is 0.


aemruli:
thanks @demboos

PR #235, opened 2017-06-21
Add ability to turn off Amazon client auto configuration.

### LINKED ISSUE #227: Please add support for conditional AWS auto configurations based on properties
There are use cases where we need to disable amazon cloud related bean auto configurations( Amazon S3, for eg). It would be nice to have auto configurations that will only activated based on a property.  `spring.cloud.aws.s3.enabled`. Similar properties can be made available to other AWS components configured by Spring Cloud.

anwarchk:
By default everything is turned on and hence all changes are backward compatible. 
 Specifying property `spring.cloud.aws.enabled=false` will turn off auto configuration
 Fixes #227

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextCredentialsAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextCredentialsAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2014 the original author or authors.
+ * Copyright 2013-2017 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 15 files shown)


artembilan:
@anwarchk ,

This is unusual and looks like very aggressive solution.

I think would be better really to drill to that your user-case when you think that you need to to disable all these auto-configurations.

That is really expectable to switch on something if you add the `-autoconfigure` jar into Spring Boot application classpath. So, having both `-autoconfigure` jar and property is something confusing overhead.

Now your turn!

anwarchk:
@artembilan If there are better way to accomplish this, I am open to suggestions. All I wanted is to not have the auto configuration run when I run unit tests or if I want to completely disable it. My idea was that, in Spring Cloud Eureka, even though you have `@EnableDiscoveryClient` on your configuration, having the property `spring.cloud.discovery.enabled=false` will disable the auto configuration. How is my approach different from that ?

artembilan:
Thank you, @anwarchk , for your feedback!

Maybe I don't know enough for Spring Boot and this Spring Cloud flexibility on top, but your arguments don't sound for me convincing.

If you deal with tests and have everything in the classpath, you don't have choice unless use `exclude` or even don't use auto-configuration scan at all, but specify those `@Configuration` classes for test config which you really are interested in.

That `spring.cloud.discovery.enabled=false` feels like a part of the whole picture. Seems for me it doesn't switch off the whole Eureka auto-config. Unlike your solutions does that for the whole SC-AWS framework. That really doesn't sound like a lever to encourage end-user to poll and confuse himself.

Yeah, that's IMO anyway...

I'm open for any other thoughts.

anwarchk:
Respect your comments @artembilan . I will go dig in further in the Spring Cloud project to see how they achieve turning off Eureka, but without really turning off the entire auto configuration. It could be the presence of a marker bean or something like that. Will update here. I will close this PR for now.

anwarchk:
Closing this PR till I figure out a less aggressive approach that would work for tests.

PR #241, opened 2017-07-10
Remove IntelliJ Files

andersonkyle:
For the purposes of making contributions easier, this PR removes IntelliJ IDE specific files from this repository.  Users no longer need to revert changes to the `.idea/` since it has been untracked and ignored.


aemruli:
Thanks for the PR @andersonkyle 
We have re-fined our IntelliJ configuration carefully in order to achieve some re-use in our project with a 
-common license header 
-inspections 
-formatter
and some other smaller project specific configuration. All other IDEA configuration files are excluded so that there should not be any changes except if people work with different IDEA version. 

I think the benefits of have some base level of conformity is good enough to keep the files and sometimes have to revert (or commit the changes). 

I will close this PR, feel free to re-open if you feel that the checked in files have a lot of changes.

Issue #207, opened 2017-03-24
Interest in support for retrieving config from EC2 Parameter Store?


wcurrie:
I recently learned about the EC2 [Parameter Store](https://aws.amazon.com/blogs/compute/managing-secrets-for-amazon-ecs-applications-using-parameter-store-and-iam-roles-for-tasks/). An example use case: distributing a database password to a spring boot app running in ECS or EC2. The password can be stored encrypted under KMS (envelope encryption). Access can be restricted based on the IAM role assigned to the container (ECS) or instance (EC2).

Seems this could be quite useful as a secret distribution mechanism for spring boot apps running on AWS. Not as elegant, feature rich and polished as Hashicorp's vault but perhaps good for simple use cases or where you don't want to run a vault cluster (backed by a consul cluster).

There's a spike with more detail in [aws-ssm-demo](https://github.com/wcurrie/aws-ssm-demo).

atmani:
Any progress on that request ? It will be helpful to have a feature where we can get encrypted value from ParameterStore.

btsteve:
Sounds like a great feature.

jkuipers:
I wrote something like this recently. The current code can be found here: https://gitlab8.trifork.nl/jorisk/spring-cloud-aws-param-store

I'd be happy to contribute it.

mrcasablr:
@jkuipers  Would you be able to convert this into a PR for the upstream folks to merge?

jkuipers:
I will have a look at what it would take to turn this into a proper pull request.

dsyer:
Closed via #308

jkuipers:
Would be nice if the updated docs could be republished as well: my code has been integrated in a republished RC2, but as there are no docs no one will know that nor how to use it.

dsyer:
That should be done now. Might take a while for caches to clear.

PR #308, opened 2018-03-23
Param store config

jkuipers:
Here's a contribution to address #207.
This code provides Spring Cloud centralized configuration support for the AWS Parameter Store by creating a PropertySourceLocator at bootstrap time that reads hierarchical configuration parameters.
Similar to Spring Cloud Config and the Consul Config support, shared configuration and overrides with profile-specific properties are supported.
Encrypted values are decrypted.

Currenty there's no support for publishing EnvironmentChangedEvents when updates are made to the parameters that make up combined PropertySources:
AFAICT the only way to that efficiently (i.e. without continuously polling all properties) would involve integrating with CloudWatch events, which I thought to be too complex for an initial implementation.

I first added my BootstrapConfiguration to the existing autoconfiguration module, but that module has a hard dependency on the context module and configures things I don' personally care for when I just want the Parameter Store integration.
That would also prevent me from running an application locally while still retrieving its config from the Parameter Store.
I therefore added autoconfiguration for this new module to its starter instead.

I've provided some documentation and a simple test: coming up with all this was already a significant amount of work, so I'd like some feedback first before putting in any more.
Please let me know what you think!


aemruli:
first of all thanks @jkuipers will have a look this week and comment on this.

SaikiranDaripelli:
@aemruli Any update on this?

docs/src/main/asciidoc/spring-cloud-aws.adoc
```diff
@@ -573,6 +585,86 @@ public class ApplicationService {
 }
 ----
 
+=== Integrating your Spring Cloud application with the AWS Parameter Store
+
+Spring Cloud provides support for centralized configuration, which can be read and made available as a regular Spring
+`PropertySource` when the application is started. The Parameter Store Configuration allows you to use this mechanism
+with the https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-paramstore.html[AWS Parameter Store].
+
+Simply add a dependency on the `pring-cloud-starter-aws-parameter-store-config` starter module to activate the support.
```
devinsba: nit: spelling

jkuipers:
Would be nice to get an update on this: BTW, I'm running (another version of) this code in production right now, so it's proving itself as I'm typing this ;)

masayuki038:
I would like this too, please.  What is the current status?

joshualo:
+1

Jarec:
+1

gamov:
This is a great addition to spring-cloud-aws, IMO.

spring-cloud-aws-parameter-store-config/src/main/java/org/springframework/cloud/aws/paramstore/AwsParamStorePropertySource.java
```diff
@@ -0,0 +1,77 @@
...
+    @Override
+    public Object getProperty(String name) {
+        return properties.get(name);
+    }
+
+    private void getParameters(GetParametersByPathRequest paramsRequest) {
+        GetParametersByPathResult paramsResult = source.getParametersByPath(paramsRequest);
+        for (Parameter parameter : paramsResult.getParameters()) {
+            String key = parameter.getName().replace(context, "").replace('/', '.');
+            properties.put(key, parameter.getValue());
```
devinsba: Might it be good to handle KMS encrypted parameters here?

jkuipers: They're already decrypted: `withWithDecryption(true)` takes care of that, assuming your service has the necessary authorities to do so.

dsyer:
Merged in 24e2f81. I also re-deployed 2.0.0.RC2 to include this commit.

jkuipers:
Thanks for merging this Dave!

mariaoks:
Is there a way to store params of application for multi tenants?

Issue #298, opened 2018-03-02
cloud.aws.region.auto ignored when cloud.aws.region.static is set


osialr:
Using Spring-cloud-aws 1.2.1 with spring-cloud-aws-autoconfigure.

The region.auto property does not have any effect on the region detection code path if a static region is set. 

We had in our `application.yml`:
```
cloud.aws.region.auto: false
cloud.aws.region.static: us-east-1
```
Then had an `application-aws.yml` profile
```
cloud.aws.region.auto: true
```

Expected Results:
Auto region detection used when run in the `aws` profile

Actual Resuts:
The region was still being forced to us-east-1 when run in the `aws` profile.

The relevant line is: https://github.com/spring-cloud/spring-cloud-aws/blob/1.2.x/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextRegionProviderAutoConfiguration.java#L42

I also tried setting the `cloud.aws.region.static` to blank but that caused an `IllegalArgumentException` within `ContextConfigurationUtils`

The only workaround was to remove the `cloud.aws.region.static` line in our `application.yml` and redo all of our yml files to default to auto detection, and include static on a per-profile basis.

petromir:
I hit the same problem.
I think that `!(this.environment.containsProperty("cloud.aws.region.static")` is redundant ( 
https://github.com/spring-cloud/spring-cloud-aws/blob/1.2.x/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextRegionProviderAutoConfiguration.java#L43).

Any plans to resolve that soon?

autocorrectoff:
Does region.auto=true work if the app is docker container?

PR #328, opened 2018-05-17
Fixes gh-298

petromir:
Check for existence of cloud.aws.region.static is change to not fail when empty string is used.


aemruli:
thanks @petromir

Issue #292, opened 2018-01-24
SQS QueueMessageHandler incompatible with Java8 Date/Time


MrBuddyCasino:
The Jackson configuration for the QueueMessageHandler cannot be influenced from the outside, and doesn't by default seem to support Java 8 Date&Time. This is not a problem for QueueMessagingTemplate, which accepts a custom MappingJackson2MessageConverter, but its hardcoded in QueueMessageHandler:

```
    private CompositeMessageConverter createPayloadArgumentCompositeConverter() {
        List<MessageConverter> payloadArgumentConverters = new ArrayList<>();

        if (JACKSON_2_PRESENT) {
            MappingJackson2MessageConverter jacksonMessageConverter = new MappingJackson2MessageConverter();
            jacksonMessageConverter.setSerializedPayloadClass(String.class);
            jacksonMessageConverter.setStrictContentTypeMatch(true);
            payloadArgumentConverters.add(jacksonMessageConverter);
        }
        [...]
    }
```
Usually the spring framework allows one to configure an ObjectMapper in a custom way, so this seems unusual. We're on version 1.2.2 btw.

Issue #264, opened 2017-09-28
SQS listeners don't take class types into account when deserializing generic objects


jtcotton63:
Say I'm writing a REST controller method that receives a generic object:

````java
@RestController
@RequestMapping("/collection")
@PreAuthorize("hasAuthority('VOLLEYMETRICS_ADMIN')")
public class SystemEventsController {

	@RequestMapping(value = "", method = RequestMethod.POST)
	public void handleItemCreatedEvent(@RequestBody CreateEvent<Item> event) {
		// Notice this works just fine
		// I can manipulate the Item to my wishes
		Item item = event.getNewValue();
		// Do stuff with the new item
	}
}
````
where the CreateEvent<T> class is:

````java
public class CreateEvent<T> {
    private final T newValue;
    private final Instant changedAt;

    public CreateEvent(@JsonProperty("newValue") T newValue, @JsonProperty("changedAt") Instant changedAt) {
        this.newValue = newValue;
        this.changedAt = changedAt;
    }

	// Getters and setters not shown
````
and the Item class is:
````java
public class Item {
    private String aString;
    private int anInt;

    // Default constructor, getters, and setters not shown
}
````

The above example works just fine. Spring is able to infer the type of the generic from the signature of the controller method. This allows Jackson to deserialize the CreateEvent.newValue property into an object of type Item.

Now, let's do the same for an SQS-listening method:

````java
@Service
@Async
public class SQSListener {

	// Autowired constructor not shown

	@SqsListener(value = "${system.property.amazon.sqs.name}", deletionPolicy = SqsMessageDeletionPolicy.ON_SUCCESS)
	public void handleItemCreatedEventViaSQS(@NotificationMessage CreateEvent<Item> event) {
        Item item = event.getNewValue();     // Exception thrown here
        // Never get to do stuff with the item
    }
}
````

This time, the an exception is thrown whenever calling event.getNewValue():

````java
java.lang.ClassCastException: java.util.LinkedHashMap cannot be cast to com.company.Item
	at com.company.service.SQSListener.handleItemCreatedEventViaSQS(SQSListener.java:48) ~[main/:na]
	at com.company.service.SQSListener$$FastClassBySpringCGLIB$$817fccac.invoke(<generated>) ~[main/:na]
	at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:204) ~[spring-core-4.3.8.RELEASE.jar:4.3.8.RELEASE]
[...]

Issue #193, opened 2017-01-09
Make it easier to customize ObjectMapper in QueueMessageHandler


andrewdmay:
We make use of custom serializers/deserializers and mixins with Jackson and need to use a customized ObjectMapper with QueueMessageHandler/@SqsListener.

We have been creating a customized QueueMessageHandlerFactory:

```
    @Bean
    public QueueMessageHandlerFactory queueMessageHandlerFactory(ObjectMapper objectMapper) {
        MappingJackson2MessageConverter mappingJackson2MessageConverter = new MappingJackson2MessageConverter();
        mappingJackson2MessageConverter.setSerializedPayloadClass(String.class);
        mappingJackson2MessageConverter.setObjectMapper(objectMapper);
        QueueMessageHandlerFactory factory = new QueueMessageHandlerFactory();
        factory.setArgumentResolvers(Arrays.asList(new PayloadArgumentResolver(mappingJackson2MessageConverter)));
        return factory;
    }
```

However while this works fine when the listener only requires the payload it does not work with the `@Header/@Headers` annotations because the PayloadArgumentResolver is added as a customer argument resolver before the standard resolvers defined in `QueueMessageHandler.initArgumentResolvers`.

I could add the header resolvers to the customer resolvers before the payload resolver, but that does not seem like an ideal solution.

Ideally if there was an `ObjectMapper` in the spring context it would be preferable if `SqsConfiguration` used it to configure the `QueueMessageHandlerFactory`.

fastnsilver:
Had to do some backflips to get this working for my purposes.  Sharing this [gist](https://gist.github.com/fastnsilver/20209f7bcdeaa5084b72a5e8857b66fd), in case anyone wants to see.  It's kind of an ugly hack because I had to employ a post processor in order to get argument resolvers to "stick" in `QueueMessageHandler`.

jtcotton63:
+1

@aemruli what is the status on this? Looks like it was assigned to someone and then they dropped it?

aemruli:
fixed with #332

PR #332, opened 2018-05-26
Add ability to customize MessageConverters for SQS listeners

maciejwalkowiak:
- since Jackson is always on the classpath it's safe to use it without conditional check
- allows to customize `MessageConverter`s by setting it on `QueueMessageHandlerFactory`
- if not set, default `MappingJackson2MessageConverter` is used
- no runtime dependency to `spring-web` anymore

Fixes: #292, #193, potentially #264


alainsahli:
Thx!

Issue #317, opened 2018-04-17
SQS Listener Container in Actuator Health Checks


pluttrell:
We are using the SQS messaging features in Spring Cloud AWS (version `1.2.1` and `2.0.0 M4`).

We had a production outage because the service(s) that support the `@SqsListener` annotation simply didn't complete configuration and logged the following message. More details can be found [here](https://github.com/spring-cloud/spring-cloud-aws/issues/316):
```
Ignoring queue with name 'exampleQueue' as it does not exist.
```

If we had a custom `HealthIndicator` (as defined [here](https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-endpoints.html#_writing_custom_healthindicators)) that could detect this error/failure state and report the service down, then our infrastructure (Kubernetes) would have restarted the app until it reported that it was healthy, thereby elimininating (or minimizing the duration of) the outage.

Is there any bean that I can inject into a custom `HealthIndicator` that I can detect this state? Or any other way to write a custom `HealthIndicator` to detect unhealthy services that are supporting the SQS messaging?

Also, perhaps a standard `HealthIndicator` should be included with the artifacts?

maciejwalkowiak:
If you mean just to check if a queue exists during runtime, you can create custom health indicator like this:

```
public class SqsQueueHealthIndicator extends AbstractHealthIndicator {

    private final AmazonSQSAsync amazonSQSAsync;
    private final String queueName;

    public SqsQueueHealthIndicator(AmazonSQSAsync amazonSQSAsync, String queueName) {
        this.amazonSQSAsync = amazonSQSAsync;
        this.queueName = queueName;
    }

    @Override
    protected void doHealthCheck(Health.Builder builder) {
        try {
            amazonSQSAsync.getQueueUrl(queueName);
            builder.up();
        } catch (QueueDoesNotExistException e) {
            builder.down(e);
        }
    }
}
```

and then use it for each queue you need:

```
@Bean
SqsQueueHealthIndicator q1HealthIndicator(AmazonSQSAsync amazonSQSAsync) {
    return new SqsQueueHealthIndicator(amazonSQSAsync, "q1");
}

@Bean
SqsQueueHealthIndicator sqsDemoHealthIndicator(AmazonSQSAsync amazonSQSAsync) {
    return new SqsQueueHealthIndicator(amazonSQSAsync, "sqs-demo");
}
```

Having such a health indicator in the library itself is a good idea.

[...]

PR #342, opened 2018-06-15
Add SqsHealthIndicator

maciejwalkowiak:
Add health indicator for SQS listeners that checks if:

- `SimpleMessageListenerContainer` is in running state for all queues mapped with `SqsListener` - even those who do not exist at startup
- all configured queues exist and are reachable.

Fixes: #317

### DIFF
--- a/spring-cloud-aws-messaging/pom.xml
+++ b/spring-cloud-aws-messaging/pom.xml
@@ -46,11 +46,6 @@
 			<artifactId>spring-messaging</artifactId>
 		</dependency>
 		<dependency>
-			<groupId>org.springframework.boot</groupId>
-			<artifactId>spring-boot-actuator</artifactId>
-			<optional>true</optional>
-		</dependency>
-		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-web</artifactId>
 			<optional>true</optional>
--- a/spring-cloud-aws-messaging/pom.xml
+++ b/spring-cloud-aws-messaging/pom.xml
@@ -46,6 +46,11 @@
 			<artifactId>spring-messaging</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-actuator</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-web</artifactId>
 			<optional>true</optional>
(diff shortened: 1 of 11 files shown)


spring-cloud-aws-messaging/pom.xml
```diff
@@ -45,6 +45,11 @@
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-messaging</artifactId>
 		</dependency>
+		<dependency>
```
aemruli: @maciejwalkowiak why do we need this dependency here ? I try to keep the boot depdencies out from the modules (only for release 2.0, later I want to add them).

maciejwalkowiak: Good point. Leftover from previous imolementation. I'll remove it soon.

maciejwalkowiak:
I am not convinced about keeping all health indicators in separate module. Would it be better to keep them close to classes they use to check health? In that case we could limit a bit the public API exposed by these classes - in this case perhaps `getMappedQueueNames` would not need to be public.

gmariotti:
Hi, is there any update on this PR?

maciejwalkowiak:
Spring Cloud GCP keeps all health indicators in `autoconfigure` module. I am thinking if in case of Spring Cloud AWS wouldn't it make more sense to also keep health indicators there and turn current module into `spring-cloud-starter-aws-cloudwatch`. What do you think @eddumelendez?

eddumelendez:
Yes, agree. IMHO, health indicators should be at the `autoconfigure` module since `spring-boot-starter-actuator` is the one that provides the initial infrastructure. So, features can be enabled just adding starters for a specific service.

Regarding to the new module `spring-cloud-starter-aws-cloudwatch`, I think make sense but something that was thinking in `spring-cloud-gcp` is that if the implementation relies on `micrometer` may  using the name`spring-cloud-starter-aws-metrics` make more sense since the `cloudwatch` configuration will be at the `autoconfigure` using mostly micrometer. It's not a big deal the final name though. WDYT?

btw, I just noticed there is no implementation in `spring-boot` but found [this comment](https://github.com/spring-projects/spring-boot/issues/15644#issuecomment-452157980) pointing to this repo.

maciejwalkowiak:
The only reason I think naming it after `cloudwatch` can be safer than `metrics` is that AWS may come up with another metrics related service and it wont be clear what we are referencing to anymore. Similar to what happened with SQS/SNS and our current `messaging` module.

Since this will be a breaking change we would either do it in 3.0 or in 2.3 (considering there will be 2.3).

maciejwalkowiak:
Closing for now, we will revisit this topic for 3.x

PR #179, opened 2016-10-25
Enables CachingConfigurer customization

venilnoronha:
This PR enables the customization of the CachingConfigurer by using ConditionalOnMissingBean(CachingConfigurer.class) over the bean. Refer #159 for more information.

I've signed the CLA. Please review and pull.

Thanks,
Venil

PR #343, opened 2018-06-15
Enable CachingConfigurer customization

### LINKED ISSUE #159: ElastiCacheCacheConfigurer bean factory should be annotated with ConditionalOnMissingBean
The user should be allowed to use his own `CachingConfigurer`, however `spring-cloud-aws` is forcing the use of `ElastiCacheCacheConfigurer`.

A solution would be to annotate the ElastiCacheCacheConfigurer factory (`ElastiCacheCachingConfiguration#cachingConfigurer`) with `ConditionalOnMissingBean(CachingConfigurer.class)`.

maciejwalkowiak:
#179 rebased against master.

Fixes #159.


maciejwalkowiak:
This adds Spring Boot dependency to the `context` module - something we prefer to avoid. `CachingConfigurer` customisation will be possible after we move all cache related configuration to `autoconfigure` module.

PR #345, opened 2018-06-18
Remove unnecessary S3 API call

cbuehler:
When creating a Resource from an ObjectSummary, we already know that the
object exists. Otherwise, we would not have an ObjectSummary about it.

So we can remove the exists() check, which causes a HEAD request for
every single matching object.

This reduces the number of HTTP calls from n+1 to 1.


aemruli:
thanks @cbuehler

PR #367, opened 2018-08-28
Support path separator for profile separation.

sbuettner:
We would like to be able to use path separators for profiles to be able to specify profile specific config under a separate path like: `/some-base-path/my-application/dev/...`


sbuettner:
Workaround that requires to hardcode the environment mapping into the `bootstrap.yml` file:
```
aws:
  paramstore:
    prefix: "/some-base-path"
    defaultContext: "my-application/dev"
```

PR #381, opened 2018-10-20
Fix broken images references

wilder:


maciejwalkowiak:
Thank you for contribution! The issue must have been resolved by some other change as all images are loaded now correctly.

Issue #359, opened 2018-07-18
@SqsListener throwing TaskRejectedException while polling AWS SQS.


hiteshbargujar:
Thanks for raising a Spring Cloud issue. What sort of issue are you raising?
Question
Please ask questions about how to use something, or to understand why something isn't
working as you expect it to, on Stack Overflow using the spring-cloud tag.

As directed I did create a question on Stack Overflow and tagged it with spring-cloud.
So far I have not received an answer :( 
Here is the link for same
https://stackoverflow.com/questions/51373082/springboot-sqslistener-not-working-with-exception-taskrejectedexception

I am trying to @SqsListener annotation for reading messages from AWS SQS.
My SQS already have 5000 messages.
As soon as I start my SpringBoot Application, I get TaskRejectedException while polling the SQS for messages.

agudian:
I stumbled across the same issue. 

If `maxNumberOfMessages` is not set, there will be a mismatch in the calculation of the threadPool size in `SimpleMessageListenerContainer#createDefaultTaskExecutor()` (which assumes that 2 worker threads will be available for each message consumer) and the `AbstractMessageListenerContainer.QueueAttributes` which will allow to fetch up to 10 messages if `maxNumberOfMessages` is not set. Messages received in one call will be scheduled for parallel execution -- so you're bound to run into this problem if more than 2 messages are actually received in one call.

Explicitly set maxNumberOfMessages on the `SimpleMessageListenerContainerFactory` bean and you should be good to go.

@aemruli in case my analysis is correct, what do you think would be the proper solution to fix the default behavior?

danieladams456:
[...]

PR #382, opened 2018-10-22
Align default number of worker threads to default SQS batch size

danieladams456:
This is a continuation of the discussion in #359.  The default number of worker threads was less than the default number of messages pulled per batch from SQS, so this resulted in `TaskRejectedException`s being thrown.

Are there any concerns around increasing the default threads to 10 to match the SQS batch size?

### DIFF
--- a/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/listener/AbstractMessageListenerContainer.java
+++ b/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/listener/AbstractMessageListenerContainer.java
@@ -58,7 +58,8 @@ abstract class AbstractMessageListenerContainer
 
 	private static final String RECEIVING_MESSAGE_ATTRIBUTES = "All";
 
-	private static final int DEFAULT_MAX_NUMBER_OF_MESSAGES = 10;
+	//Visible to child classes for thread pool sizing
+	protected static final int DEFAULT_MAX_NUMBER_OF_MESSAGES = 10;
 
 	private static final int DEFAULT_WAIT_TIME_IN_SECONDS = 20;
 
--- a/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/listener/SimpleMessageListenerContainer.java
+++ b/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/listener/SimpleMessageListenerContainer.java
@@ -201,7 +201,7 @@ public class SimpleMessageListenerContainer extends AbstractMessageListenerConta
 					.setCorePoolSize(spinningThreads * DEFAULT_WORKER_THREADS);
 
 			int maxNumberOfMessagePerBatch = getMaxNumberOfMessages() != null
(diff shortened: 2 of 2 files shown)


JanGroot:
Came here to fix just this :-) so +1 from me.

In my project, as a fix, we set the maxmessages to 4 , so batchsize would be 4 and max threads would be (4 + 1) times queues. We chose 4 as 10 seemed a bit high for our use case. Maybe queue depth in the taskexecutor would be a property to consider, but then again it would be more prudent to just configure a taskexecutor explicitly anyway. 

Was a bit nasty as well this, with multiple queues configured all seems ok, when load hits one queue there are enough worker threads (due to the ‘spinning threads’ multiplier. But when load arrives on all queues, errors all over the place :-/ 

Again, just on the alignment of the batchsize and max threads to prevent the errors +1 :-)

danieladams456:
Good call @JanGroot.  Does this look right to you?  I defaulted [`maxNumberOfMessages`](https://github.com/spring-cloud/spring-cloud-aws/blob/9d0eec31ef56cf0f865d51ece693a8cc2525a8a5/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/listener/AbstractMessageListenerContainer.java#L70) in the abstract class so it should flow down to the [max thread pool calculation](https://github.com/spring-cloud/spring-cloud-aws/blob/3bfd3a537192cf38a14c7a0c8df16876511db42a/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/listener/SimpleMessageListenerContainer.java#L186-L187) and dropped the default max threads change.

danieladams456:
I see that broke the test: `MessageListenerContainerTest.testMaxNumberOfMessages:150 expected null, but was:<10>`
Is this the right way to fix it though?

danieladams456:
Sounds good.  What do you think of this in the child class?  Keep it?  The child class shouldn't have to assume the parent always will return a non null value?
```java
int maxNumberOfMessagePerBatch = getMaxNumberOfMessages() != null ? getMaxNumberOfMessages() : DEFAULT_WORKER_THREADS;
```

JanGroot:
In Holland we have a saying, there are multiple roads to Rome. 
You could also just change the visibility of the DEFAULT_MAX_NUMBER_OF_MESSAGES and just reference that in the child class

danieladams456:
Good idea.  That keeps less consumer logic in the abstract class.  Going forward with that.

StephanPraetsch:
We ran into the same problem. The workaround described [here](https://github.com/spring-cloud/spring-cloud-aws/issues/359) works. Nevertheless I'm looking forward to seeing this fixed in the lib. Is there any progress here? Right now I see conflicting files.

raoofm:
@ryanjbaxter  @OlgaMaciaszek fyi

ryanjbaxter:
@danieladams456 could you fix the merge conflicts?

danieladams456:
@ryanjbaxter looking at it now.  I'll verify that this change logic is still valid since the files have diverged.  On first glance, it just seems like a file-wide formatting change is causing the conflict.

danieladams456:
Rebased on the latest master, then fixed tabs vs spaces. 😄 @ryanjbaxter should be good to merge now.
Thanks!

PR #390, opened 2018-11-16
Make some S3 related classes and methods public

tgianos:
`SimpleStorageResource` was package private making it difficult to use outside of this library. Making this public allows users to use or extend this convenient `Resource` implementation in ways the library doesn't currently support.

`AmazonS3ProxyFactory` was a public class but the factory method was package private. Making this public lets others leverage the convienent AOP logic contained within for S3 clients.


spencergibb:
Thanks, @tgianos for your first contribution to spring-cloud-aws. Looking forward to more!

PR #392, opened 2018-11-28
Customizable aws client configuration for cloudwatch

fred84:
AWSClient for sending metrics to cloudwatch has default configuration. For example default thread pool size in AWSClient is 50, which in my case it is only used to send metrics. I think it is reasonable to have an option to customize this configuration.

I've only updated configuration for cloudwatch. There is similar MailSenderAutoConfiguration.


maciejwalkowiak:
Thanks for submitting the PR. We applied similar approach to all integrations in https://github.com/spring-cloud/spring-cloud-aws/pull/725

PR #293, opened 2018-01-29
Support primitive types for SQS DataType

joshuaali:
As well as supporting primitive types, this also makes the optional custom type just that: optional.

Resolves #221.

PR #397, opened 2018-12-13
Fix bugs in SQS header parsing for Number types, and add unit tests.

### LINKED ISSUE #374: SqsListener String index out of bounds issue
I'm encountering a problem when trying to use the @SQSListener annotation from the Spring Cloud module. My Listener method is defined as follows:
```
@SqsListener(value = "myproject-dev-au-error-queue")
public void listenPhoenix(String message) throws IOException {

    logger.info(message);
 }

```
And when I start receiving messages from the queue i get the following stack trace:

```
Exception in thread "simpleMessageListenerContainer-4" Exception in thread "simpleMessageListenerContainer-6" Exception in thread "simpleMessageListenerContainer-9" Exception in thread "simpleMessageListenerContainer-10" java.lang.StringIndexOutOfBoundsException: String index out of range: -1
    at java.lang.String.substring(String.java:1931)
    at org.springframework.cloud.aws.messaging.core.QueueMessageUtils.getNumberValue(QueueMessageUtils.java:93)
    at org.springframework.cloud.aws.messaging.core.QueueMessageUtils.getMessageAttributesAsMessageHeaders(QueueMessageUtils.java:80)
    at org.springframework.cloud.aws.messaging.core.QueueMessageUtils.createMessage(QueueMessageUtils.java:56)
    at org.springframework.cloud.aws.messaging.listener.SimpleMessageListenerContainer$MessageExecutor.getMessageForExecution(SimpleMessageListenerContainer.java:375)
    at org.springframework.cloud.aws.messaging.listener.SimpleMessageListenerContainer$MessageExecutor.run(SimpleMessageListenerContainer.java:336)
    at org.springframework.cloud.aws.messaging.listener.SimpleMessageListenerContainer$SignalExecutingRunnable.run(SimpleMessageListenerContainer.java:392)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
    at java.lang.Thread.run(Thread.java:745)
```

With the problem being in QueueMessageUtils.getNumberValue(...), more specifically

```
private static Object getNumberValue(MessageAttributeValue value) {
    String numberType = value.getDataType().substring("Number".length() + 1);
    ....
}
```
This is happening on all versions of Spring Cloud (Dalston, Finchley, Edgeware, etc.)

Any ideas on what might be causing this behaviour?

kevinhwang:
Fixes #374.

### DIFF
--- a/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/core/QueueMessageUtils.java
+++ b/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/core/QueueMessageUtils.java
@@ -21,13 +21,23 @@ import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageHeaders;
 import org.springframework.messaging.MessagingException;
 import org.springframework.messaging.support.GenericMessage;
+import org.springframework.util.ClassUtils;
 import org.springframework.util.MimeType;
 import org.springframework.util.NumberUtils;
 
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.AbstractMap;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.UUID;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+import static java.util.stream.Collectors.collectingAndThen;
+import static java.util.stream.Collectors.toMap;
 
 /**
  * @author Alain Sahli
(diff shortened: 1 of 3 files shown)


spring-cloud-aws-messaging/pom.xml
```diff
@@ -55,6 +55,11 @@
 			<artifactId>spring-webmvc</artifactId>
 			<optional>true</optional>
 		</dependency>
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-lang3</artifactId>
```
spencergibb: Please remove this dependency and reimplement.

kevinhwang: Sure. Take a look now. Also added support for non-qualified wrapper types (e.g., `Number.Integer` works like `Number.int`—no need for `Number.java.lang.Integer`).

kevinhwang: @spencergibb Any comments?

spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/core/QueueMessageUtils.java
```diff
@@ -38,6 +48,23 @@
     private static final String RECEIPT_HANDLE_MESSAGE_ATTRIBUTE_NAME = "ReceiptHandle";
     private static final String MESSAGE_ID_MESSAGE_ATTRIBUTE_NAME = "MessageId";
 
+    private static final Map<Class<? extends Number>, Class<? extends Number>> PRIMITIVE_NUMBER_TYPE_TO_WRAPPED_TYPE_MAP = Stream.of(
```
spencergibb: I didn't know about this before, but checkout `org.springframework.util.NumberUtils`. Looks like it will handle this without much custom code.

kevinhwang: `org.springframework.util.NumberUtils#parseNumber(String, Class)` unfortunately doesn't support primitive types (e.g., `int.class`), only the wrapped types—it'll throw an exception.

That's why I added a map to look up the wrapped type given a primitive type (and unfortunately there doesn't seem to be a simple to do it without relying on Apache Commons or Google Guava or other libraries that provided a primitive -> wrapped type util method).

We could eliminate a majority of this look up stuff if we just don't support stuff like `"Number.int"`, etc., and only permit the wrapped types, e.g., `"Number.Integer"`. What do you think?

kevinhwang: @spencergibb Any update on this? What would you prefer?

dberrueta-atlassian: Note there's another open PR (#293 ) to add support for primitive types.

maciejwalkowiak:
Thanks @kevinhwang for your PR - from now on we will try to tackle PRs on time. The issue is fixed now.

PR #460, opened 2019-03-29
Fix duplicate error logging

sayembd:
During the processing of an exception thrown by a queue message handler method, the AbstractMethodMessageHandler always logs the error [when there is no existing handler method for that exception](https://github.com/spring-projects/spring-framework/blob/f273fa990c5231684b0dedf4895d9e6bd8f66235/spring-messaging/src/main/java/org/springframework/messaging/handler/invocation/AbstractMethodMessageHandler.java#L592). On the other hand, the SimpleMessageListenerContainer logs the same exception again [when the message deletion policy is set to ON_SUCCESS](https://github.com/spring-cloud/spring-cloud-aws/blob/8d456c9428c930fe44f2d00c6144a1be6354770e/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/listener/SimpleMessageListenerContainer.java#L440). This results in duplicated error logging.

This commit fixes this issue by delegating exception processing to AbstractMethodMessageHandler only when there is a configured exception handler.


sayembd:
I have also signed the Pivotal CLA.

PR #524, opened 2020-02-14
Upgrade to AWS SDK 2.0

tinexw:
Upgraded AWS SDK to version 2.0 according to the [changelog](https://github.com/aws/aws-sdk-java-v2/blob/master/docs/LaunchChangelog.md). 

There are some potential blockers where I would like to get some feedback on:

# Potential Blockers
## 1. AmazonSQSBufferedAsyncClient not supported
The [AmazonSQSBufferedAsyncClient](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/sqs/buffered/AmazonSQSBufferedAsyncClient.html) is no longer supported according to the [AWS docs](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-client-side-buffering-request-batching.html):

> The AWS SDK for Java 2.x isn't currently compatible with the AmazonSQSBufferedAsyncClient. 

There already is a ticket in the AWS SDK GitHub but it's still open: https://github.com/aws/aws-sdk-java-v2/issues/848 resp. https://github.com/aws/aws-sdk-java-v2/issues/165

Not really sure what to do... I mean as a last resort I could copy all the classes from the old SDK and update them I'd rather not...

## 2. No common interface for Sync and Async client
The 1.x SDK had a common interface for the sync and async client. This is no longer the case. 

Previously, the user could configure either the async or the sync client and pass it to certain framework classes, e.g. `NotificationMessagingTemplate`. This won't be possible anymore. As a first step I would suggest to only accept one client (either sync or async - depending on which methods are used) in all customizable classes. As a next step, the classes could be duplicated to accept both clients (if needed).

Initially I thought that it was also previously possible to replace the autoconfigured sync client with the async client. But I think that was not possible because of the way the [ConditionalOnMissingAmazonClient](https://github.com/spring-cloud/spring-cloud-aws/blob/master/spring-cloud-aws-context/src/main/java/org/springframework/cloud/aws/context/annotation/ConditionalOnMissingAmazonClient.java) works. Can you confirm this @alainsahli ?

## 3. Messaging: Sync and Async client used
The messaging module currently uses the async client and calls sync and async methods on it. This is no longer possible in the new SDK. You need to use the sync client for sync methods and the async client for async methods. I see the following options with the new SDK:
- Keep functionality the same by adding both clients and using the respective one (I implemented it like this for now)
- Use async methods everywhere
- Use sync methods everywhere

Please let me know what you think @spencergibb @OlgaMaciaszek 

# Things to do
- [x] Upgrade libraries and fix compile errors
- [ ] Discuss and resolve potential blockers
  - [ ] AmazonSQSBufferedAsyncClient
  - [ ] No common interface
  - [ ] Both clients used in messaging
- [x] Resolve other remaining TODOs in code (non straightforward compile errors)
- [x] Run and fix all unit tests
  - [x] core
  - [x] context
  - [x] jdbc
  - [x] messaging
  - [x] autoconfigure
- [ ] Integration Tests
  - [x] Upgrade libraries and adapt to new SDK
  - [ ] Run and fix all issues that are found
- [ ] Manual testing of everything not covered by integration tests
- [ ] Review changelog if there are more changes in addition to the obvious ones that need to be considered
- [ ] (Optional) Prettify code e.g. rename `describeStacksResult` variable to `describeStacksResponse` (all `Result` class names were changed to `Response`)
- [ ] Update documentation

### DIFF
--- a/spring-cloud-aws-actuator/pom.xml
+++ b/spring-cloud-aws-actuator/pom.xml
@@ -30,13 +30,13 @@
 	<description>Spring Cloud AWS Actuator</description>
 
 	<properties>
-		<micrometer.version>1.0.4</micrometer.version>
+		<micrometer.version>1.3.3</micrometer.version>
 	</properties>
 
 	<dependencies>
 		<dependency>
 			<groupId>io.micrometer</groupId>
-			<artifactId>micrometer-registry-cloudwatch</artifactId>
+			<artifactId>micrometer-registry-cloudwatch2</artifactId>
 			<version>${micrometer.version}</version>
 		</dependency>
 		<dependency>
--- a/spring-cloud-aws-autoconfigure/pom.xml
+++ b/spring-cloud-aws-autoconfigure/pom.xml
@@ -63,13 +63,13 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
-			<groupId>com.amazonaws</groupId>
-			<artifactId>aws-java-sdk-ses</artifactId>
+			<groupId>software.amazon.awssdk</groupId>
+			<artifactId>ses</artifactId>
 			<optional>true</optional>
 		</dependency>
 		<dependency>
-			<groupId>com.amazonaws</groupId>
-			<artifactId>aws-java-sdk-cloudwatch</artifactId>
+			<groupId>software.amazon.awssdk</groupId>
+			<artifactId>cloudwatch</artifactId>
 			<optional>true</optional>
 		</dependency>
 		<dependency>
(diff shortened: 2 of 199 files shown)


smaldini:
Wow - that's quite a contribution ! We will have a look on our side as well when possible.

maciejwalkowiak:
Thanks @tinexw for your efforts. Even though this PR won't be merged we will use it as a reference when moving existing functionality to AWS SDK v2 in Spring Cloud AWS 3.0 and make sure to give you credits.

tinexw:
Thanks @maciejwalkowiak for letting me know! At least I have some closure now ;-) Let me know if you need further support for Spring Cloud AWS 3.0!

PR #525, opened 2020-02-17
Adding support for EventBridge

jmnarloch:
Summary: Adds support for AWS EventBridge to the `spring-cloud-aws-messaging`. EventBridge is service launched in August 2019. EventBridge allows to decouple the service applications by publishing events, to which application/services can subscribe to.

To some degree the integration is very similar to SNS integration, currently EventBridge allows only publishing the events. Consuming them is possible through AWS Lambda.

Compared to SNS though there are couple differences:
* Events published to EventBus does not have headers. If there is need to support headers or message attributes through EventBridge the headers needs to be pass as event payload.
* There is no notion of Message ID in EventBus on client side.
* Each event is identified by pair of `source` and `detail-type`.

Reference:
* https://docs.aws.amazon.com/eventbridge/latest/userguide/what-is-amazon-eventbridge.html


pivotal-issuemaster:
@jmnarloch Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-aws&pullRequestId=525)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-aws&pullRequestId=525) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@jmnarloch Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-aws&pullRequestId=525)!

.idea/misc.xml
```diff
@@ -1,13 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="JavaScriptSettings">
```
jmnarloch: It would probably makes sense to remove the IDEA specific files from the repository.

spring-cloud-aws-dependencies/pom.xml
```diff
@@ -31,7 +31,7 @@
 	<name>Spring Cloud AWS Dependencies</name>
 	<description>Spring Cloud AWS Dependencies</description>
 	<properties>
-		<aws-java-sdk.version>1.11.415</aws-java-sdk.version>
+		<aws-java-sdk.version>1.11.624</aws-java-sdk.version>
```
jmnarloch: The changes that are part of this Pull Request requires the SDK from mid 2019.

jmnarloch:
As a side question, is there any other package that would require to be changed to add full support for AWS Events? How about Spring Cloud Streams?

.idea/misc.xml
jmnarloch: I remove the file in latest commit.

jmnarloch:
@marcingrzejszczak Thanks for reviewing the change, I address the comment and pushed updated revision.

jmnarloch:
@marcingrzejszczak any other feedback?

jmnarloch:
@spencergibb would you mind taking a look.

spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/config/annotation/EnableEventBridge.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.context.annotation.Import;
+
+/**
+ * @author Jakub Narloch
+ * @since 2.3.0
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Import({ EventBridgeConfiguration.class })
+public @interface EnableEventBridge {
```
spencergibb, line 33: Spring Cloud has moved away from `@Enable` annotations for the most part. Seems like this could just be auto-configuration with `@ConditionalOnClass` guards

Issue #406, opened 2019-01-29
Configuration properties compatible with Spring Boot Relaxed Binding


lukasznyrc:
> Spring Boot uses some relaxed rules for binding Environment properties to @ConfigurationProperties beans, so there does not need to be an exact match between the Environment property name and the bean property name.
> Common examples where this is useful include dash-separated environment properties (for example, **context-path** binds to **contextPath**).

This is not a valid case when it comes to Spring Cloud AWS configuration, for example it's not possible to use kebab case, which is recommended for use in .properties and .yml files. Following configuration doesn't work.

```yml
cloud:
  aws:
    credentials:
      use-default-aws-credentials-chain: true
      instance-profile: false
      profile-name: my-profile
```
A working configuration has to be in camel case, which is inconsistent with Spring Boot recommendations:
```yml
cloud:
  aws:
    credentials:
      useDefaultAwsCredentialsChain: true
      instanceProfile: false
      profileName: my-profile
```

Spring Boot: 2.1.2.RELEASE
Spring Cloud AWS: 2.1.0.RELEASE (Greenwich.RELEASE)

internetstaff:
I just hit this myself on V2.1.1.RELEASE.

PR seems like it would break non-kebab binding, but that does highlight the source of the problem:

In ContextCredentialsAutoConfiguration, AwsCredentialsProperties has a @ConfigurationProperties annotation for "cloud.aws.credentials". 

IntelliJ picks this up and suggests cloud.aws.credentials.use-default-aws-credentials-chain for a property in a yaml config.

It looks like registerBeanDefinitions() then does direct lookups of several properties by property names, avoiding binding and meaning only the camel case version of "cloud.aws.credentials.useDefaultAwsCredentialsChain" and other properties works.

Perhaps AwsCredentialsProperties needs to be extracted then autowired into ContextCredentialsAutoConfiguration for use instead of directly querying properties?

lukasznyrc:
> PR seems like it would break non-kebab binding
No. It works fine for both naming conventions.

> Perhaps AwsCredentialsProperties needs to be extracted then autowired into ContextCredentialsAutoConfiguration for use instead of directly querying properties?

What do you mean exactly?

internetstaff:
I am surprised that camel case would still work when explicitly querying snake case, since the current problem seems to be that snake case doesn't work because it's explicitly querying for camel case. :)

[...]

PR #407, opened 2019-01-29
Rename config properties in ContextCredentialsAutoConfiguration

lukasznyrc:
Rename config properties in ContextCredentialsAutoConfiguration to kebab case to enable Spring Boot relaxed binding - resolves gh #406

PR #536, opened 2020-03-28
Issue 406 - children of property "cloud.aws.credentials" supporting Relaxed Binding.

tac-yacht:
Fixes gh-406
This fix has Backward compatible.


maciejwalkowiak:
Thanks @tac-yacht for contribution. #407 solves the same issue in a bit different way. I am taking the production code from #407 and tests from your PR.

PR #552, opened 2020-05-07
Fix parameterstore to be not eager

mbazhutkin-at-trading-point-com:
* Added onlyApplicationNameLocator to handle cases when we want to use only application name without environment setup
* Added check if application name and default value for locator path are same not to duplicate requests to SSM

### DIFF
--- a/spring-cloud-aws-parameter-store-config/src/main/java/org/springframework/cloud/aws/paramstore/AwsParamStoreProperties.java
+++ b/spring-cloud-aws-parameter-store-config/src/main/java/org/springframework/cloud/aws/paramstore/AwsParamStoreProperties.java
@@ -66,6 +66,9 @@ public class AwsParamStoreProperties {
 	/** Is AWS Parameter Store support enabled. */
 	private boolean enabled = true;
 
+	/** Is use only application name for resource location*/
+	private boolean onlyApplicationNameLocator = true;
+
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@mbazhutkin-at-trading-point-com Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-aws&pullRequestId=552)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-aws&pullRequestId=552) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@mbazhutkin-at-trading-point-com Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-aws&pullRequestId=552)!

maciejwalkowiak:
Thanks for contribution! I like the idea but I would flip the property. Instead of `onlyApplicationNameLocator` I believe its simpler to reason if we have `includeProfiles`. This by default must be set to `true` for backward compatibility. Please also include tests.

maciejwalkowiak:
Superseded by #596

Issue #553, opened 2020-05-08
Fargate Region Support


dejonghe:
**Is your feature request related to a problem? Please describe.**
When running in Fargate, EC2 Metadata is not available, which causes looking up the region to fail. Based on the implementation of getRegion() or getCurrentRegion() this is nothing this is not a bug of spring-cloud-aws. Note the "If not in EC2 return null" in the following doc. 
https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/regions/Regions.html

In Fargate, EC2 Metadata is not available however, the environment variable AWS_REGION is, and is also set on most local testing environments.

**Describe the solution you'd like**
In the following function of spring-cloud-aws test if the returned value of getRegion() is null and if so check for the environment variable AWS_REGION and use that. 

https://github.com/spring-cloud/spring-cloud-aws/blob/master/spring-cloud-aws-core/src/main/java/org/springframework/cloud/aws/core/region/RegionProvider.java

**Describe alternatives you've considered**
Currently the work around is to set the property of "static region" as follows;
`cloud.aws.region.static=${AWS_REGION}`

**Additional context**
I looked for others that might have raise this issue but did not find thing. Basically this caused a bit of a run around when trying to deploy to Fargate, and I think by implimenting this small change it would save many people quite a bit of hassle if / when they deploy to Fargate.

giova333:
@maciejwalkowiak this pr should fix the problem https://github.com/spring-cloud/spring-cloud-aws/pull/559

PR #559, opened 2020-05-31
Issue 553: AWS region should be retrieved considering multiple strategies.

giova333:
Currently, spring cloud aws uses `Ec2MetadataRegionProvider` for dynamic region lookup.
However, this solution doesn't work in some cases such https://github.com/spring-cloud/spring-cloud-aws/issues/553
We should consider using  `com.amazonaws.regions.DefaultAwsRegionProviderChain` that provides multiple strategies for the region lookup

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/test/java/org/springframework/cloud/aws/autoconfigure/context/ContextRegionProviderAutoConfigurationTest.java
+++ b/spring-cloud-aws-autoconfigure/src/test/java/org/springframework/cloud/aws/autoconfigure/context/ContextRegionProviderAutoConfigurationTest.java
@@ -22,7 +22,7 @@ import org.junit.After;
 import org.junit.Test;
 
 import org.springframework.boot.test.util.TestPropertyValues;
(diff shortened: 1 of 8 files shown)


maciejwalkowiak:
Thanks @giova333 for PR. I am working on the same issue at exactly same time with a bit different approach as we cannot break backward compatibility at this stage. I appreciate your efforts - give me some time to mark issues as "ideal-to-contribution" and then you could pick them up.

giova333:
@maciejwalkowiak Sure, no problem.

maciejwalkowiak:
@giova333 this is a change I had in mind #560

giova333:
@maciejwalkowiak Looks good. The idea to enable `DefaultAwsRegionProviderChain` via property is great. I think in the major release it should be enabled by default.

maciejwalkowiak:
Yes I believe so too 👍 Either in 2.3 or 3.0 default chains for both credentials and regions will be used unless static values are provided.

Issue #519, opened 2020-01-06
nit: Typo


econn9:
**Describe the bug**
[incorrectly spelled word in Sysout output
](https://github.com/spring-cloud/spring-cloud-aws/blob/a1fa62f577fdf3e01e6c1137d88e8c484a09466c/spring-cloud-aws-integration-test/src/test/java/org/springframework/cloud/aws/messaging/AbstractContainerTest.java#L50)
> throw new Exception("Couldn't stop container withing 15 seconds");

tmnuwan12:
How do we take on fixing bugs. Some one get assigned to work on it..?

maciejwalkowiak:
I'll go through all the issues again in upcoming day(s) and mark ones that can be picked up with "help wanted". There is also some contributor guidelines needed as not everything is obvious - like for example not breaking backward compatibility for 2.2.x branch. If you would like to fix this one it would be great 👍

tmnuwan12:
Yes I will fix this one. Waiting for further instruction from @maciejwalkowiak

maciejwalkowiak:
@tmnuwan12 I went through issues. Check ones with "help wanted" label: https://github.com/spring-cloud/spring-cloud-aws/issues?q=is%3Aopen+is%3Aissue+label%3A%22help+wanted%22

tmnuwan12:
@maciejwalkowiak is there any ci pipeline I can use to test my changes or we have to use our aws credentials when running integration tests..?

maciejwalkowiak:
Once you create pull request there will be a build executed against your branch. It won't execute tests connecting to AWS though. I think there is a space for improvement there but I'm not sure yet how to tackle it.

PR #561, opened 2020-05-31
Fix typo in the exception text

tmnuwan12:
Fixes gh-519


maciejwalkowiak:
Thanks for PR! Surprisingly CI does not run against pull requests. We will need to change it.

Issue #361, opened 2018-07-23
Ability to get s3:// URI back from an S3 resource


osi:
I am trying to "round-trip" resources from a s3:// URI into a `Resource` and back again. Currently, `SimpleStorageResource#getURL` returns a `https://` URI into AWS. 

It would be nice to be able to round-trip these URIs

tmnuwan12:
I want to help. Where should I look into.?

maciejwalkowiak:
@tmnuwan12 go for it. Just keep in mind that whatever we do here should stay backward compatible.

tmnuwan12:
Can you assign it to me.. But I need some time with this if that's ok. I have some office stuff to clear out  this week.

tmnuwan12:
For bug fixes do we need to branch off from master or 2.2.x branch..?

maciejwalkowiak:
Good question! Ideally from 2.2.x

tmnuwan12:
I checked the code for SimpleStorageResource. From what I understand from the requirement is ability to specify "s3://" endpoint when creating SimpleStorageResource and return 
 "s3://" compliant URL by calling method getURL().. Before I change anything want to check the approach here.. I was thinking of add/modify constructor to accommodate a URL object or url string and return a compatible URL based on constructor argument.. Feel free to provide some inputs please.

tmnuwan12:
@maciejwalkowiak

maciejwalkowiak:
I think it it should be possible to obtain s3:// and https:// no matter how what parameters were passed to the constructor. What do you think?

osi:
agreed @maciejwalkowiak

tmnuwan12:
@osi @maciejwalkowiak Yes I agree. What do you think about getURI() implementation for SimpleStorageResource class that provides the "s3://" URI.

tmnuwan12:
@maciejwalkowiak  I am getting some compilation errors when running  maven tests ( but in Intellij editor no compilation errors). Is this something known issue or am I missing something..?

[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.0:testCompile (default-testCompile) on project spring-cloud-aws-core: Compilation failure: Compilation failure:
[21,17] package org.junit does not exist

maciejwalkowiak:
[...]

PR #565, opened 2020-06-06
Add aws s3:// uri support

tmnuwan12:
Fixes gh-361

### DIFF
--- a/spring-cloud-aws-core/src/main/java/org/springframework/cloud/aws/core/io/s3/SimpleStorageResource.java
+++ b/spring-cloud-aws-core/src/main/java/org/springframework/cloud/aws/core/io/s3/SimpleStorageResource.java
@@ -26,6 +26,8 @@ import java.io.OutputStream;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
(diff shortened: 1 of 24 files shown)


spring-cloud-aws-core/src/main/java/org/springframework/cloud/aws/core/io/s3/SimpleStorageResource.java
```diff
@@ -143,6 +145,10 @@ public URL getURL() throws IOException {
 				"/" + this.bucketName + "/" + this.objectName);
 	}
 
+	public URI getS3URI() throws URISyntaxException {
```
maciejwalkowiak, line 148: Lets change method name to `getS3Uri` - when you look at other classes in Spring Framework whenever they have `URI` in name its spelled `Uri`.

`URISyntaxException` is rather impossible to happen right? We should rather catch it and throw `new RuntimeException` with explanation, similar to how it's done in `AmazonS3ClientFactory`.

spring-cloud-aws-core/src/main/java/org/springframework/cloud/aws/core/io/s3/SimpleStorageResource.java
```diff
@@ -143,6 +145,10 @@ public URL getURL() throws IOException {
...
+		return new URI("S3", "//" + this.bucketName + "/" + this.objectName, null);
```
maciejwalkowiak, line 149: `S3` should be rather lowercase `s3`

spring-cloud-aws-core/src/main/java/org/springframework/cloud/aws/core/io/s3/SimpleStorageResource.java
tmnuwan12: Ok. Fixed as requested.

spring-cloud-aws-core/src/main/java/org/springframework/cloud/aws/core/io/s3/SimpleStorageResource.java
tmnuwan12: Ok. Fixed as requested.

tmnuwan12:
> Compilation issues is a result of Spring Boot just recently dropping JUnit 4. I'll add very soon JUnit 4 to dependencies both on `master` and `2.2.x` branches so make sure to rebase.

I have rebased against master.

maciejwalkowiak:
Thanks @tmnuwan12! Your contribution has been merged.

PR #569, opened 2020-06-06
Add auto-configuration that adds EC2 metadata as tags for the Microme…

renanreismartins:
Add auto-configuration that adds EC2 metadata as tags for the Micrometer metrics.

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/micrometer/AwsEc2MicrometerMetricsAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/micrometer/AwsEc2MicrometerMetricsAutoConfiguration.java
@@ -73,13 +73,6 @@ public class AwsEc2MicrometerMetricsAutoConfiguration {
 	}
 
 	Collection<Tag> createTags() {
-		// Java 9
-		// Collection<Tag> tags = metricsProperties.getTags()
-		// .stream()
-		// .map(this::retrieveMetadataValue)
-		// .flatMap(Optional::stream)
-		// .collect(toList());
-
 		return metricsProperties.getTags()
 								.stream()
 								.map(this::retrieveMetadataValue)
--- a/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/micrometer/AwsEc2MicrometerMetricsAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/micrometer/AwsEc2MicrometerMetricsAutoConfiguration.java
@@ -1,0 +1,94 @@
+/*
+ * Copyright 2013-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.aws.autoconfigure.micrometer;
+
+import java.util.Collection;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+import io.micrometer.core.instrument.ImmutableTag;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+
+import org.springframework.boot.actuate.autoconfigure.metrics.CompositeMeterRegistryAutoConfiguration;
+import org.springframework.boot.actuate.autoconfigure.metrics.MeterRegistryCustomizer;
+import org.springframework.boot.actuate.autoconfigure.metrics.MetricsAutoConfiguration;
+import org.springframework.boot.actuate.autoconfigure.metrics.export.simple.SimpleMetricsExportAutoConfiguration;
+import org.springframework.boot.autoconfigure.AutoConfigureAfter;
+import org.springframework.boot.autoconfigure.AutoConfigureBefore;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.cloud.aws.autoconfigure.context.ContextInstanceDataAutoConfiguration;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.env.Environment;
+
+import static java.util.Optional.ofNullable;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Configuration that adds EC2 metadata as tags for the Micrometer metrics.
+ *
+ * @author Renan Reis Martins de Paula
+ */
+@Configuration(proxyBeanMethods = false)
+@EnableConfigurationProperties(AwsEc2MicrometerMetricsProperties.class)
+@AutoConfigureBefore({ CompositeMeterRegistryAutoConfiguration.class,
+		SimpleMetricsExportAutoConfiguration.class })
+@AutoConfigureAfter({ MetricsAutoConfiguration.class,
+		ContextInstanceDataAutoConfiguration.class })
+@ConditionalOnClass(MeterRegistry.class)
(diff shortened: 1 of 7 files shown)


renanreismartins:
@maciejwalkowiak Did some research and came up with this pull request. Also run it on an app demo and got the expected results.

The build complains about junit in all modules, however I haven't changed any dependency related code. Could you help me to find what is missing?

maciejwalkowiak:
Please rebase against master. Spring Boot dropped JUnit 4 in latest snapshots and this is where compilation issues are coming from.

spring-cloud-aws-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -18,11 +18,16 @@
       "description": "Cloud watch namespace.",
       "type": "java.lang.String"
     },
+    {
```
maciejwalkowiak, line 21: There is no need to add it to `additional-spring-configuration-metadata` as the metadata will be generated from `@ConfigurationProperties` class.

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/micrometer/AwsEc2MicrometerMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,94 @@
...
+ */
+@Configuration(proxyBeanMethods = false)
+@EnableConfigurationProperties(AwsEc2MicrometerMetricsProperties.class)
+@AutoConfigureBefore({ CompositeMeterRegistryAutoConfiguration.class,
+		SimpleMetricsExportAutoConfiguration.class })
+@AutoConfigureAfter({ MetricsAutoConfiguration.class,
+		ContextInstanceDataAutoConfiguration.class })
+@ConditionalOnClass(MeterRegistry.class)
+@ConditionalOnProperty(prefix = "cloud.aws.ec2.micrometer.metrics", name = "tags")
+public class AwsEc2MicrometerMetricsAutoConfiguration {
```
maciejwalkowiak, line 57: I would prefer to move new bean to `CloudWatchExportAutoConfiguration`. Do you think it's worth having separate auto-configuration class?

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/micrometer/AwsEc2MicrometerMetricsProperties.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Collection;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+import static java.util.Collections.EMPTY_LIST;
+
+/**
+ * @author Renan Reis Martins de Paula
+ */
+@ConfigurationProperties(prefix = "cloud.aws.ec2.micrometer.metrics")
```
maciejwalkowiak, line 28: Are these metrics specific to EC2? Would they will be also sent when running on ECS or EKS? I think the configuration properties prefix should be `cloud.aws.cloud-watch`.

maciejwalkowiak, line 28: Please add Javadocs to both class and field (not to getters & setters). Javadocs will be used in configuration metadata.

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/micrometer/AwsEc2MicrometerMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,94 @@
...
+		this.environment = environment;
+	}
+
+	@Bean
+	public MeterRegistryCustomizer meterRegistryCustomizer() {
+		return registry -> registry.config().commonTags(createTags());
+	}
+
+	Collection<Tag> createTags() {
+		// Java 9
```
maciejwalkowiak, line 76: Please remove this commented code - we have to stick to Java 8.

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/micrometer/AwsEc2MicrometerMetricsAutoConfiguration.java
renanreismartins: It is a good idea, CloudWatchExportAutoConfiguration is enable specifically by the management.metrics.export.cloudwatch. This auto-configuration explicitly tells they are independent and the tags can still be enabled without cloudwatch. Makes sense?

spring-cloud-aws-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
renanreismartins: Good point, thanks. Reverted.

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/micrometer/AwsEc2MicrometerMetricsAutoConfiguration.java
renanreismartins: Done

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/micrometer/AwsEc2MicrometerMetricsProperties.java
renanreismartins: done.

renanreismartins: Since the properties are injected in the Environment by AmazonEc2InstanceDataPropertySource I assumed it is specifically for EC2.

renanreismartins:
> I did not even realise that it will be just about taking values from the environment. Since it's not specific to CloudWatch it could have been a general purpose `MeterRegistryCustomizer`, like `EnvironmentVariablesMeterRegistryCustomizer`, potentially added to Spring Boot. What do you think @eddumelendez?

I also found out later that getting the values from the Environment would be a more straight forward solution. Still not related to CloudWatch, however still related to AWS, that is why looks like it makes sense to be on the aws module. But please, let me know if you guys have a different opinion.

eddumelendez:
> I did not even realise that it will be just about taking values from the environment.

In this case looks like is due to properties are loaded by `AmazonEc2InstanceDataPropertySource`. But, if they are available as a system environments will be loaded by `spring-boot` too. 

> Since it's not specific to CloudWatch it could have been a general purpose MeterRegistryCustomizer

Yes, I like the idea to customize but I think it will be general purpose I guess. So, `MeterRegistryCustomizer` will be better I think.

renanreismartins:
Hi @eddumelendez, thanks for the feedback.

> In this case looks like is due to properties are loaded by `AmazonEc2InstanceDataPropertySource`. But, if they are available as a system environments will be loaded by `spring-boot` too.

Yes, the props are loaded by AmazonEc2InstanceDataPropertySource.

Regarding of how the variables are available, when I mentioned Environment I referred to Spring Environment. They are not available as system env variables.

renanreismartins:
Hi @eddumelendez @maciejwalkowiak, how should we proceed here?

maciejwalkowiak:
Thank you @renanreismartins for your patience. We needed some time to digest it. We've decided that this should not be a part of Spring Cloud AWS as fetching properties from environment - as much as handy it is - is not Spring Cloud AWS specific. I think you can consider proposing this change to Spring Boot project. For now I believe it will be enough for Spring Cloud AWS to update reference documentation on how to add EC2 metadata to Micrometer using custom `MeterRegistryCustomizer`. Is this something you would like to do?

renanreismartins:
Hi @maciejwalkowiak, at first glance it looks like as just getting the properties from the environment. There is more to it, the purpose is to retrieve AWS metadata and add to the metrics, it just turned that the metadata is already in the environment.

Currently, every developer that needs the metadata in the metrics tags, has to write an `MeterRegistryCustomizer` and inject the metadata there. To balance both points of view, I agree that it is not a big deal, my point is that it is a very common requirement so we could just provide the option of configuring on the properties what metadata they wanted on the metrics tag.

Saving the developer to create the custom `MeterRegistryCustomizer`, creating property class and wiring everything up.

If after taking it in consideration you still think we should go with the documentation, I will add it after the section
**Using instance tags** (https://github.com/spring-cloud/spring-cloud-aws/blob/master/docs/src/main/asciidoc/spring-cloud-aws.adoc#using-instance-tags)

Please let me know what is your final decision.

maciejwalkowiak:
Apologies @renanreismartins for keeping you waiting so long. In the end we've decided to not provide this customizer out of the box.

Issue #496, opened 2019-10-03
stackResourceRegistryFactoryBean returns null value


reypader:
Encountering a problem with starting an application using `spring-cloud-starter-aws:2.1.0.RELEASE` with `cloud.aws.stack.auto = false`. The application encounters the following error:  

```Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'stackResourceRegistryFactoryBean' defined in class path resource [org/springframework/cloud/aws/autoconfigure/context/ContextStackAutoConfiguration.class]: Initialization of bean failed; nested exception is java.lang.ClassCastException: org.springframework.beans.factory.support.NullBean cannot be cast to org.springframework.beans.factory.FactoryBean```

This is likely due to `org.springframework.cloud.aws.autoconfigure.context.ContextStackAutoConfiguration.stackResourceRegistryFactoryBean()` returning null.

```java
@Bean
    @ConditionalOnMissingBean(StackResourceRegistry.class)
    public StackResourceRegistryFactoryBean stackResourceRegistryFactoryBean(AmazonCloudFormation amazonCloudFormation) {

        if (StringUtils.hasText(environment.getProperty("cloud.aws.stack.name"))) {
            return new StackResourceRegistryFactoryBean(amazonCloudFormation, new StaticStackNameProvider(this.environment.getProperty("cloud.aws.stack.name")));
        }

        if (environment.getProperty("cloud.aws.stack.auto") == null || "true".equalsIgnoreCase(environment.getProperty("cloud.aws.stack.auto"))) {
            return new StackResourceRegistryFactoryBean(amazonCloudFormation, new AutoDetectingStackNameProvider(amazonCloudFormation, this.amazonEC2));
        }

        return null;
    }
```

Should this be changed to use `@ConditionalOnProperty("cloud.aws.stack.name")` and `@ConditionalOnProperty("cloud.aws.stack.auto")` ?

JanGroot:
same issue in my project.

maciejwalkowiak:
I am not able to reproduce it using 2.2.1 version. Could you please provide complete sample as a github repository that reproduces this issue?

adannunez:
Running into the same issue. Just adding some more info that might be help troubleshoot.

[...]

PR #574, opened 2020-06-07
Fix StackResourceRegistry creation when CloudFormation support is turned off

maciejwalkowiak:
Fixes gh-496

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextStackAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextStackAutoConfiguration.java
@@ -22,10 +22,10 @@ import com.amazonaws.services.cloudformation.AmazonCloudFormationClient;
 import com.amazonaws.services.ec2.AmazonEC2;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnSingleCandidate;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.aws.autoconfigure.context.properties.AwsStackProperties;
 import org.springframework.cloud.aws.context.annotation.ConditionalOnMissingAmazonClient;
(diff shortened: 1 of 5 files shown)


maciejwalkowiak:
@eddumelendez I will appreciate your review.

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextStackAutoConfiguration.java
```diff
@@ -60,24 +65,29 @@
...
-			AmazonCloudFormation amazonCloudFormation) {
-
-		if (StringUtils.hasText(environment.getProperty("cloud.aws.stack.name"))) {
-			return new StackResourceRegistryFactoryBean(amazonCloudFormation,
-					new StaticStackNameProvider(
-							this.environment.getProperty("cloud.aws.stack.name")));
-		}
+	@ConditionalOnMissingBean
+	@ConditionalOnProperty("cloud.aws.stack.name")
+	public StackNameProvider staticStackNameProvider() {
```
eddumelendez, line 70: it is the expected behaviour that `manual` setting has a priority over the default `auto-detection` but I think we are getting the expected result due to the order of the beans. I would prefer to be explicit about it. In this case,

```
if (StringUtils.hasText(this.awsStackProperties.getName)) {
    return new StaticStackNameProvider(awsStackProperties.getName());
}
else {
    return new AutoDetectingStackNameProvider(amazonCloudFormation, this.amazonEC2);
}
```

Also, we can delete `auto` property. WDYT?

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextStackAutoConfiguration.java
```diff
@@ -22,33 +22,38 @@
...
 @Configuration(proxyBeanMethods = false)
 @Import({ ContextCredentialsAutoConfiguration.class,
 		ContextDefaultConfigurationRegistrar.class })
 @ConditionalOnClass(name = "com.amazonaws.services.cloudformation.AmazonCloudFormation")
+@EnableConfigurationProperties(AwsStackProperties.class)
 public class ContextStackAutoConfiguration {
 
 	@Autowired
-	private Environment environment;
+	private AwsStackProperties awsStackProperties;
```
eddumelendez, line 56: just `properties`?

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextStackAutoConfiguration.java
maciejwalkowiak: Indeed we are getting correct behaviour because of the order of beans. Here the problem is that if we change it the way you suggested we are changing the behaviour of the method. I still do want to change how stack autoconfiguration works and potentially combine `auto` and `stackName` into single property but this can happen in 2.3.x. This bugfix is meant for 2.2.x

eddumelendez: I was thinking about it. Sounds good to me then :)

maciejwalkowiak:
Merged in 69628d22382ff47817af23a52cc7935510c9b46b

PR #581, opened 2020-06-09
Enforce spring-javaformat plugin

eddumelendez:


eddumelendez:
@maciejwalkowiak do you agree to enforce the format? I think we can benefit from it when contributions are made. Also, I have been thinking about targeting `2.2.x`. WDYT?

PR #582, opened 2020-06-09
Add property key to enable/disable services integration

eddumelendez:


eddumelendez:
it's pending to add tests

maciejwalkowiak:
Looks good! I think in the future we would add there properties to corresponding `@ConfigurationProperties` classes as there is a plan anyway to add ability to configure custom endpoint - so that they would go away from `additional-configuration-properties`, but this can be done once we get there.

eddumelendez:
most of the properties that are declared in `@ConditionalOnProperty` are not declared in `@ConfigurationProperties` because they are not use anywhere. Are you planning to used these properties as part of the logic?

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextStackAutoConfiguration.java
```diff
@@ -51,6 +51,7 @@
 		ContextDefaultConfigurationRegistrar.class })
 @ConditionalOnClass(name = "com.amazonaws.services.cloudformation.AmazonCloudFormation")
 @EnableConfigurationProperties(AwsStackProperties.class)
+@ConditionalOnProperty(name = "cloud.aws.stack.enabled", havingValue = "true")
```
maciejwalkowiak, line 54: This is one way we can tackle disabling Cloudformation but I am wondering why to even add Cloudformation dependency in first place. 

Wouldn't it make more sense if we mark Cloudformation dependency as optional and enable configuration if user adds it to the classpath? I am thinking about similar approach to other integrations provided in context module.

eddumelendez: agree on that. I will remove this commit in order to be address the issue later then.

PR #583, opened 2020-06-09
Use default AWS SDK credentials and region providers by default

maciejwalkowiak:
Unless other credentials provider configured, Spring Cloud AWS will use now default AWS region chain. Similar in region auto-configuration.

If user decides to opt-out from defaults, all credentials provider have to be now explicitly configured, there are no default values that I believe makes it easier to reason about what is happening under the hood.

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextCredentialsAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextCredentialsAutoConfiguration.java
@@ -34,6 +34,7 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.cloud.aws.autoconfigure.context.properties.AwsCredentialsProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.util.StringUtils;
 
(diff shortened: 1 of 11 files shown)


spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextCredentialsAutoConfiguration.java
```diff
@@ -16,100 +16,69 @@
...
+	@ConditionalOnMissingBean(name = CREDENTIALS_PROVIDER_BEAN_NAME)
+	public AWSCredentialsProvider awsCredentialsProvider(
+			AwsCredentialsProperties properties) {
 
-	/**
-	 * Registrar for the credentials provider.
-	 */
-	public static class Registrar
-			implements ImportBeanDefinitionRegistrar, EnvironmentAware {
+		List<AWSCredentialsProvider> providers = new ArrayList<>();
```
eddumelendez, line 57: what about having this `List<AWSCredentialsProvider> providers = resolveCredentialsProvider(properties);` and the method `private List<AWSCredentialsProvider> resolveCredentialsProvider(properties)` contains all the logic

maciejwalkowiak: 👍

Issue #566, opened 2020-06-06
Migrate to JUnit 5


maciejwalkowiak:
Spring Boot dropped automatic support for JUnit 4 (https://github.com/spring-projects/spring-boot/commit/8fb7a6cace5caeeb0c7177bc3acd3915e24e7648). We should also move to JUnit 5 as it is currently a de facto standard for testing Java code.

Issue #572, opened 2020-06-07
Introduce Testcontainers & LocalStack to run integration


rieckpil:
**Is your feature request related to a problem?** 

- Running the current integration tests requires spawning resources in AWS 
- The setup for this requires some manual effort (see [README](https://github.com/spring-cloud/spring-cloud-aws/blob/master/README.adoc))
-  Spring Cloud AWS does not execute the integration tests by default

**Describe the solution you'd like**

With the help of [Testcontainers](https://www.testcontainers.org/modules/localstack/) and [LocalStack](https://github.com/localstack/localstack) we can mirror AWS resources locally using Docker containers and save $ 

**Describe alternatives you've considered**

Simplify the AWS resource setup or ask AWS for some free budget for running the integration test.

**Things to consider**

- Not sure if there are some corner cases for AWS resources that LocalStack does not support

rieckpil:
I could start working on a first POC and maybe convert a first _old_ integration test to this _new_ setup. 

Some questions I would have for this:

- Should we still put them in a own Maven module `spring-cloud-aws-integration-test` package or make them part of each specific module (e.g. SQS integration tests inside the messaging module)
- In regards to https://github.com/spring-cloud/spring-cloud-aws/issues/566 I guess new tests should use JUnit 5

maciejwalkowiak:
Thanks @rieckpil for taking care of it! I believe there will be a still small number of tests that will have to touch real AWS environment - I would keep them in in `spring-cloud-aws-integration-test` module. The rest as you suggested - SQS integration tests go to messaging module. 

Let's write new tests in JUnit 5 only double check please that both are executed by surefire plugin.

eddumelendez:
@rieckpil you can use branch `2.3.x`

DarrenForsythe:
Regarding JUnit5, I could take a swing at moving the project over to the JUnit5 Jupiter APIs. There doesn't appear to be anything stopping it from an initial attempt. Some `ExpectedException` rules which could easily be replaced by `Assertations#assertThatCode`

rieckpil:
For me it's okay. I am not touching any existing tests and rather create a new one, so this shouldn't clash

tmnuwan12:
It would nice if you can write some doc about it share it also. Great take away from this effort can be applied to elsewhere in other projects in the industry IMO. Just a suggestion give time constraints.

PR #587, opened 2020-06-09
Junit 5 upgrade

DarrenForsythe:
* Remove JUnit 4 jar
* Exclude transitive JUnit 4 dependency from Spring Boot Starter Test
* Remove all JUnit imports and replace with their JUnit 5 Jupiter replacements
* Remove all JUnit 4 ExpectedException, and `Test(expected = ...)`.  Replace with AssertJ assertThatThrownBy
* Replace RunWith with ExtendWith and the runners with their JUnit 5 extenions.

Fixes gh-566


eddumelendez:
thanks @DarrenForsythe ! I think would great to include removing `public` for classes and test too.

DarrenForsythe:
@eddumelendez wansn't sure if that would have been too much for one PR, but I did a sweep and removed all public modifiers in test classes where appropriate

DarrenForsythe:
I noted there is 30 checkstyle issues within the integration test module, I know that new ITs are currently being written but I could provide a separate PR to fix those as looking at gh-572 those ITs will still exist?

maciejwalkowiak:
Thanks @DarrenForsythe for this massive PR! IT are under development, hard to say right now which parts of existing module stays and which will be gone.

maciejwalkowiak:
Merged in 2.3.x, soon also in master :-)

PR #588, opened 2020-06-10
Introduce Testcontainers & LocalStack for the first integration tests

rieckpil:
This relates to https://github.com/spring-cloud/spring-cloud-aws/issues/572. 

I started working on the integration test for `spring-cloud-aws-messaging` and wanted to move them to the `spring-cloud-aws-messaging` project instead of `spring-cloud-aws-integration-test`. To make them work the module needs a dependency on `spring-cloud-aws-autoconfigure` to test e.g. the Spring Boot integration. This results in a cyclic dependency between autoconfigure and the messaging module as now both rely on each other. 

Should I rather stick to leaving the integration test in `spring-cloud-aws-integration-test` or can you think of any other solution to make this work?

Any help appreciated 👍


rieckpil:
As I couldn't find a solution for the cyclic dependency I decided to **not** move the current integration test out of the `spring-cloud-aws-integration-test` module and rather make them work with LocalStack.

Right now all IT inside the `message` Java package of `spring-cloud-aws-integration-test` run successfully using LocalStack and Testcontainers.

Some noteworthy adjustments I made:

- Remove `@EnableAutoConfiguration` for the Boot tests in favor of `@ImportAutoConfiguration({MessagingAutoConfiguration.class}`. Otherwise too much would try  to autoconfigure (EC2, Database, etc.) itself and would require a full-blown setup
- I create the SNS topic and all SQS queues with Java using Testcontainers `.execInContainer()` inside `@BeforeAll`. This can be refactored to use a `.sh` script which includes all required `aws` cli commands to prepare the LocalStack containers. I had issues in the past with copying a executable file in the past (https://stackoverflow.com/questions/61258421/how-to-copy-executable-file-to-docker-container-using-testcontainers) but can have a look at it again

Next, as the IT are now still in the `spring-cloud-aws-integration-test` module, I would propose to include this module to the parent project but ensure only the migrated integration tests are executed on every build.

.circleci/config.yml
```diff
@@ -37,4 +37,4 @@ jobs:
           destination: artifacts
       - store_test_results:
           path: ~/junit/
-          destination: testartifacts
\ No newline at end of file
+          destination: testartifacts
```
eddumelendez, line 40: I think we should revert the change in this file

rieckpil: good catch, thank you

rieckpil:
@eddumelendez should I squash the commits to two commits? One for Testcontainers and one for formatting & AssertJ?

eddumelendez:
@rieckpil that would be great

rieckpil:
@eddumelendez done 👍

eddumelendez:
Thank you so much @rieckpil

rieckpil:
if you are fine with the setup, I can continue with the other integration tests, after this was merged 👍

maciejwalkowiak:
Thanks @rieckpil for PR and apologies for delayed reply. We need to wait till the end of this week to see to what degree we will be able to test on actual AWS cloud and what goes to Localstack. We will keep you updated!

rieckpil:
@maciejwalkowiak any updates on this?

maciejwalkowiak:
Apologies @rieckpil, we fucked up with this one. For now we decided to stick to running tests against real AWS services. We will revisit Localstack topic in 3.0.

rieckpil:
no worries @maciejwalkowiak :)

PR #593, opened 2020-06-12
Move OnConditionalAwsCloudEnvironment to autoconfigure module

eddumelendez:


eddumelendez:
I was thinking on moving `AwsCloudEnvironmentCheckUtils` too.

@maciejwalkowiak WDYT?

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/condition/ConditionalOnAwsCloudEnvironment.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.context.annotation.Conditional;
+
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Conditional(OnAwsCloudEnvironmentCondition.class)
+public @interface ConditionalOnAwsCloudEnvironment {
```
maciejwalkowiak, line 31: Please add Javadocs. I know that many classes have zero Javadocs but I think we should raise the bar from now on :-)

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/condition/OnAwsCloudEnvironmentCondition.java
```diff
@@ -0,0 +1,39 @@
...
+package org.springframework.cloud.aws.autoconfigure.condition;
+
+import org.springframework.boot.autoconfigure.condition.ConditionOutcome;
+import org.springframework.boot.autoconfigure.condition.SpringBootCondition;
+import org.springframework.cloud.aws.context.support.env.AwsCloudEnvironmentCheckUtils;
+import org.springframework.context.annotation.ConditionContext;
+import org.springframework.core.type.AnnotatedTypeMetadata;
+
+/**
+ * @author Eddú Meléndez
```
maciejwalkowiak, line 26: Since this class is not completely new but is based on old one, we should give credits to original authors too.

maciejwalkowiak:
Merged in ec784c4b8ae12a6282633f185b5187420e9ac4b2

Issue #589, opened 2020-06-10
Create a Test Compile step for Spring Cloud AWS Integration Test in Circle CI 


DarrenForsythe:
**Is your feature request related to a problem? Please describe.**
Currently the the `spring-cloud-aws-integration-test` module is not defined as a module in the projects root pom, and manually needs to have a the `test-compile` phase ran on it to have some confidence that there are no checkstyle issues, and no compilation errors.

**Describe the solution you'd like**
While new integration tests are being proposed in gh-572 this module may be of value to continue to allow full E2E tests with AWS that Localstack may not be able to replicate, or produce a failing test. 

A solution would be to at least include the module in a Circle CI step to do a `test-compile`. This would at least catch any compilation and checkstyle/format issues. 

**Describe alternatives you've considered**
The module could be included in parent POM with the tests re-categorized to only execute with a given profile, or test annotation type. This would allow the maven reactor to at least execute every phase, minus the ITs unless specified. Keeping the project in sync and alleviating IDE issues or other gotchas

**Additional context**
N/A

eddumelendez:
> While new integration tests are being proposed in gh-572 this module may be of value to continue to allow full E2E tests with AWS that Localstack may not be able to replicate, or produce a failing test.

Let's see how that works against `localstack`. But, totally agree we should enable this in the CI

I think we can use profiles so by default running against `localstack` using `testcontainers` and with a profile `aws-integration` against real aws instances.

/cc @maciejwalkowiak WDYT?

DarrenForsythe:
There's a small PR that would solve this using JUnit5 Tag and an extra Maven profile that will exclude aws integration tests from running, while still compiling checkstyle etc as the module is now part of the reactor.

maciejwalkowiak:
Thanks @DarrenForsythe! Fixed in https://github.com/spring-cloud/spring-cloud-aws/commit/f920fc35d4b4371ad9ea0c28ba5d6b7fa80518e7

PR #597, opened 2020-06-13
Add integration module as first class support

DarrenForsythe:
fixes gh-589


rieckpil:
as part of moving the first integration test to Testcontainers, I also fixed formatting and move the IT to `assertThat`: https://github.com/spring-cloud/spring-cloud-aws/pull/588 PR is currently pending

maciejwalkowiak:
Thanks @DarrenForsythe for contribution. I totally agree to run compilation & checkstyle by default. With this PR we have two profiles: `integration-test` and `aws-integration-test`. Since `integration-test` module is included by default, I guess we could just drop `integration-test` profile?

DarrenForsythe:
Failed to see the first profile sitting there, I've removed it in favour of `aws-integration-test` as I think it gives more context to what the profile does vs. having ITs generally run as part of the plan

eddumelendez:
just 2 things:

1. every file modified should be updated to 2020 in the license header. Not a big deal we can do it in the merge process.
2. branch 2.2.x is not using junit5

eddumelendez:
Just noticed that the `@AWSIntegration` annotation is in a base test classes at `spring-cloud-aws-integration-test` module which makes me wonder if should have that annotation at all. May just allowing to run the module with a specific profile is enough. Am I missing something?

DarrenForsythe:
@eddumelendez this will be incompatiable with the 2.2.x branch as is due to the `Tag`, it could be retrofited for 2.2.x and add the JUnit4 `Category` too

eddumelendez:
@DarrenForsythe I was looking into the `Category` but since all the tests depend on `@RunWith(SpringRunner.class)` already I think would be hard to do.  WDYT about https://github.com/spring-cloud/spring-cloud-aws/pull/597#issuecomment-656423285 ?

DarrenForsythe:
The reasoning for the profile was  given issue #588 there is a need to seperate tests that can run locally vs. tests that need an AWS Account. From experience localstack is great, but there can be cases you want a E2E test using an AWS Account, also the build would fail without exlcuding the AWS Tests currently. 

Options in my mind, 



* I _think_ we could try adding the Junit4 annotations to the meta `AWSIntegration` e.g. include JUnit4 as an optional depdenencyh and configure surefire for each? 
* Disable all AWS Integration tests to be re-written with localstack, are the AWS Integration tests being ran at the minute? 
* AWS Open Source Credits? that would limit the tests to CI Still but could activate on CI through other configuration

eddumelendez:
> I think we could try adding the Junit4 annotations to the meta AWSIntegration e.g. include JUnit4 as an optional dependency and configure surefire for each?

The best approach to me and I just did an example so we will be fine. TBH, never did that with inheritance so TIL 😃 . I will apply those changes in the merge process.

eddumelendez:
@DarrenForsythe thank you so much for your contribution. PR is merged in master via f920fc35d4b4371ad9ea0c28ba5d6b7fa80518e7

Issue #539, opened 2020-04-05
Warn or throw Exception if SQS Message couldn't be deleted


GabrielBB:
By default, messages should be deleted if the reading and processing was successful (Listener method didn't throw error). However, if the queue doesn't have `RemovePermission` on AWS, no warn or error is thrown, making it look like everything happened ok. 

```java
@SpringBootApplication
public class DemoApplication {

    public static void main(String[] args) {
        SpringApplication.run(DemoApplication.class, args);
    }

    @SqsListener(value = "test-queue")
    public void queueListener(String message) {
        System.out.println(message);
    }
}
```

![image](https://user-images.githubusercontent.com/8752823/78505622-f3e4ef00-7774-11ea-9d88-d1fbf98c8b22.png)

maciejwalkowiak:
Indeed the result of the delete message is ignored. We should at least log an error when it happens.

https://github.com/spring-cloud/spring-cloud-aws/blob/5ac4cdbf7b98a032ba7205d2c5d6054b6e0ca604/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/listener/SimpleMessageListenerContainer.java#L446-L449

karmanov:
hi @maciejwalkowiak, can I take this one?

maciejwalkowiak:
@karmanov go for it 👍

maciejwalkowiak:
@karmanov do you think you will be able to implement it in foreseeable future?

karmanov:
hi @maciejwalkowiak yes, I will implement it this week.

PR #600, opened 2020-06-17
Log an error if SQS Message couldn't be deleted

mkatircioglu:
Fixes #gh-539

### DIFF
--- a/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/listener/QueueMessageAcknowledgment.java
+++ b/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/listener/QueueMessageAcknowledgment.java
@@ -18,8 +18,12 @@ package org.springframework.cloud.aws.messaging.listener;
 
 import java.util.concurrent.Future;
 
+import com.amazonaws.handlers.AsyncHandler;
 import com.amazonaws.services.sqs.AmazonSQSAsync;
 import com.amazonaws.services.sqs.model.DeleteMessageRequest;
+import com.amazonaws.services.sqs.model.DeleteMessageResult;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * @author Alain Sahli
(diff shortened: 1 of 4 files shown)


spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/listener/QueueMessageAcknowledgment.java
```diff
@@ -33,6 +37,8 @@
 
 	private final String receiptHandle;
 
+	private final Logger logger = LoggerFactory.getLogger(getClass());
```
maciejwalkowiak, line 40: Logger could be `static`

spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/listener/SimpleMessageListenerContainer.java
```diff
@@ -440,7 +442,23 @@ private void applyDeletionPolicyOnError(String receiptHandle) {
 
 		private void deleteMessage(String receiptHandle) {
 			getAmazonSqs().deleteMessageAsync(
-					new DeleteMessageRequest(this.queueUrl, receiptHandle));
+					new DeleteMessageRequest(this.queueUrl, receiptHandle),
```
maciejwalkowiak, line 445: Since this handler is exactly the same as in previous class (for a good reason), wouldn't it make sense to extract it to an actual class instead of anonymous class?

spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/listener/QueueMessageAcknowledgment.java
mkatircioglu: Done

spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/listener/SimpleMessageListenerContainer.java
mkatircioglu: Done

PR #602, opened 2020-06-18
AwsParamStorePropertySourceLocator gets extra contexts

### LINKED ISSUE #596: AwsParamStorePropertySourceLocator gets extra contexts
**Describe the bug**
The list of contexts is duplicated.

**Sample**
Given the following configuration:

> aws:
>  paramstore:
>       enabled: true
>       prefix: ''
>       profile-separator: '/'
>       default-context: ${spring.application.name}
>       fail-fast: false

I can see four contexts: 

Screen Shot 2020-06-12 at 5 51 36 PM

The contexts size should be 2 instead of 4.

MatejNedic:
Fixes #596 
I noticed that AwsSecretsManagerPropertySourceLocator creates duplicate contexts as well so I implemented fix and tests scenarios for that as well.


spring-cloud-aws-parameter-store-config/src/main/java/org/springframework/cloud/aws/paramstore/AwsParamStorePropertySourceLocator.java
```diff
@@ -85,9 +86,11 @@ public AwsParamStorePropertySourceLocator(AWSSimpleSystemsManagement ssmClient,
 		this.contexts.add(defaultContext + "/");
 		addProfiles(this.contexts, defaultContext, profiles);
 
-		String baseContext = prefix + "/" + appName;
-		this.contexts.add(baseContext + "/");
-		addProfiles(this.contexts, baseContext, profiles);
+		if (!this.properties.getDefaultContext().equals(appName)) {
```
maciejwalkowiak, line 89: Have you considered changing `this.context` from `List` to `TreeSet` so that it guarantees element uniqueness without adding any `ifs` later in the code?

MatejNedic: Will change.

MatejNedic:
Hi @maciejwalkowiak ,
since we are using Junit 5 in 2.3.x should we refactor these tests for 2.3.x but with usage of @ParametaraizedTest ?  As well in 2.3.x we are using private builders for properties in property test I think we should extract them out in public class since both tests in this PR are using same class builders. Do you think we should do that?

spring-cloud-aws-parameter-store-config/src/main/java/org/springframework/cloud/aws/paramstore/AwsParamStorePropertySourceLocator.java
```diff
@@ -92,8 +92,6 @@ public AwsParamStorePropertySourceLocator(AWSSimpleSystemsManagement ssmClient,
 			addProfiles(this.contexts, baseContext, profiles);
 		}
 
-		Collections.reverse(this.contexts);
```
maciejwalkowiak, line 92: There's been a missing test that justifies this line. Context have to ordered in a way that the most specific ones are first - profiles, application, then the default context. Would be nice if you could include such test and then I think it should fail without reversing the `context` collection

MatejNedic: Makes sense, will implement test for both Locator classes.

maciejwalkowiak:
> Hi @maciejwalkowiak ,
> since we are using Junit 5 in 2.3.x should we refactor these tests for 2.3.x to use@ParametaraizedTest instead ? As well in 2.3.x we are using private builders for properties in property test I think we should extract them out in public class since both tests in this PR are using same class builders. Do you think we should do that?

I will take care of merging these changes in 2.3.x that will include refactoring as you mentioned. In 2.2.x lets stick to JUnit 4.

spring-cloud-aws-parameter-store-config/src/main/java/org/springframework/cloud/aws/paramstore/AwsParamStorePropertySourceLocator.java
MatejNedic: @maciejwalkowiak since TreeSet wouldn't allow order of profiles,application and then the default context because String comparing I implemented it with LinkedHashSet to preserve order and remove duplicates.

.idea/misc.xml
```diff
@@ -10,4 +10,4 @@
   <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" project-jdk-name="1.8" project-jdk-type="JavaSDK">
     <output url="file://$PROJECT_DIR$/classes" />
   </component>
-</project>
\ No newline at end of file
+</project>
```
eddumelendez, line 13: avoid including this change

spring-cloud-aws-parameter-store-config/src/main/java/org/springframework/cloud/aws/paramstore/AwsParamStorePropertySourceLocator.java
```diff
@@ -32,6 +32,7 @@
 import org.springframework.core.env.PropertySource;
 import org.springframework.util.ReflectionUtils;
 
+
```
eddumelendez, line 35: remove empty line

.idea/misc.xml
MatejNedic: Done

spring-cloud-aws-parameter-store-config/src/main/java/org/springframework/cloud/aws/paramstore/AwsParamStorePropertySourceLocator.java
MatejNedic: Done

maciejwalkowiak:
Thanks @MatejNedic. Since it changes type of `contexts` fields I'll merge it to 2.3.x branch. Unlikely that someone relies on this field being a list but still, 2.2.x must not contain breaking changes.

spring-cloud-aws-parameter-store-config/src/test/java/org/springframework/cloud/aws/paramstore/AwsParamStorePropertySourceLocatorTest.java
```diff
@@ -0,0 +1,141 @@
...
+		AwsParamStorePropertySourceLocator locator = new AwsParamStorePropertySourceLocator(
+			ssmClient, properties);
+		env.setActiveProfiles("test");
+		locator.locate(env);
+
+
+		List contextToBeTested =  new ArrayList(locator.getContexts());
+
+		assertThat(contextToBeTested.get(0)).isEqualTo("application/messaging-service_test/");
+		assertThat(contextToBeTested.get(1)).isEqualTo("application/application_test/");
```
maciejwalkowiak, line 101: If you run this test against current version without your changes, the order will be different:

1. application specific profiles
2. application
3. default profiles
4. default

So index `1` is flipped with `2`.

spring-cloud-aws-parameter-store-config/src/main/java/org/springframework/cloud/aws/paramstore/AwsParamStorePropertySourceLocator.java
```diff
@@ -82,14 +83,14 @@ public AwsParamStorePropertySourceLocator(AWSSimpleSystemsManagement ssmClient,
 		String prefix = this.properties.getPrefix();
 
 		String defaultContext = prefix + "/" + this.properties.getDefaultContext();
-		this.contexts.add(defaultContext + "/");
+		String appContext = prefix + "/" + appName;
+
+		addProfiles(this.contexts, appContext, profiles);
```
maciejwalkowiak, line 88: Look first at the comment in the unit test. To match the order we maintain currently I think this should be more like:

```
addProfiles(this.contexts, appContext, profiles);
this.contexts.add(appContext + "/");

addProfiles(this.contexts, defaultContext, profiles);
this.contexts.add(defaultContext + "/");
```

It could be also that I am missing something so please correct me if I am wrong.

spring-cloud-aws-parameter-store-config/src/test/java/org/springframework/cloud/aws/paramstore/AwsParamStorePropertySourceLocatorTest.java
MatejNedic: I saw same behaviour as well, but since you wrote that profiles should go first I understood that both profiles should be followed by application name. Sorry for missunderstanding. Should I change the order?

MatejNedic: I saw same behaviour as well, but since you wrote that profiles should go first I understood that both profiles should be followed by application name. Sorry for missunderstanding. Should I change the order?

maciejwalkowiak:
I did little polishing and made `getContext` return `List` so that it can go in `2.2.x`. Thanks @MatejNedic for another contribution! Merged in 582fdc6691ead7904aaed176af2cf85221edacd5

Issue #554, opened 2020-05-13
Issue with Profile Separator


salah-java:
The Amazon AWS has possibility of adding '/' as profile separator.
I have secret names configured at AWS Secret Manager as
'/secret/application/configuration'

when I set '/' as profile separator, Bean validation triggers & application boot up fails.

Following are the logs

```
Binding to target org.springframework.boot.context.properties.bind.BindException: Failed to bind properties under 'aws.secretsmanager' to org.springframework.cloud.aws.secretsmanager.AwsSecretsManagerProperties failed:

    Property: aws.secretsmanager.profileSeparator
    Value: /
    Reason: must match "[a-zA-Z0-9.\-_]+"

```

Please add the '/' to allowed list of profile separators

giova333:
I think that this pr should fix the issue https://github.com/spring-cloud/spring-cloud-aws/pull/466/commits/3336c98179c1d7a5d49cbb71af136e6374c8cfaa

maciejwalkowiak:
I am wondering if there is any reason why we shouldn't allow `/` as profile separator in both Secrets Manager and Parameter Store integrations.

Issue #512, opened 2019-12-06
Support for / as a profile separator


Girija-Shankar-Swain:
**Problem**
Now the default profile separator is `_`.  If an application name is `some-spring-boot-app` the paramstore path becomes `\config\some-spring-boot-app_{profile}\{propertyName}` 

The most general use case for spring profiles is to have multiple environment specific configurations e.g. `dev`, `qa`, `prod` etc.  So the valid config paths would look like following

`\config\some-spring-boot-app_dev\{property-name}`
`\config\some-spring-boot-app_qa\{property-name}`
`\config\some-spring-boot-app_prod\{property-name}`

Now we also use paramstore to store config for non-spring based projects and where the standard is `\config\{app-name}\{environment-name}\{property-name}`
It is a little frustrating to maintain config property name space in different ways.

**Solution**
It would be great if we also support `/` as spring profile separator. The current [validation ](https://github.com/spring-cloud/spring-cloud-aws/blob/e36b4fd495203f0c28c07f08fc2954e7a2bbdc05/spring-cloud-aws-parameter-store-config/src/main/java/org/springframework/cloud/aws/paramstore/AwsParamStoreProperties.java#L47)does not allow this.

**Describe alternatives you've considered**
 I do not have the context why the regex pattern does not allow this. If there is a compelling reason why it should not be supported then we live with what it supports, otherwise would be happy to send pull request.

PR #603, opened 2020-06-19
Allow `/` and `\` as profile separator in Secrets Manager and Parameter Store integrations

maciejwalkowiak:
Fixes gh-554
Fixes gh-512

### DIFF
--- a/spring-cloud-aws-parameter-store-config/pom.xml
+++ b/spring-cloud-aws-parameter-store-config/pom.xml
@@ -62,5 +62,17 @@
 			<optional>true</optional>
 		</dependency>
 
+		<dependency>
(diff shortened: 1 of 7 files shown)


spring-cloud-aws-parameter-store-config/pom.xml
```diff
@@ -62,5 +62,17 @@
 			<optional>true</optional>
 		</dependency>
 
+		<dependency>
```
eddumelendez, line 65: I am not sure about bringing these dependencies for the validation

maciejwalkowiak: They have test scope and since in 2.3 we moved away from validation API, in 2.3 they will be removed

PR #609, opened 2020-06-19
Update copyright in all files from 2019 to 2020.

maciejwalkowiak:


maciejwalkowiak:
I know this one is not very exciting to review

eddumelendez:
I just wonder if all files were modified this year

maciejwalkowiak:
Oh, then I think I misunderstood idea of this copyright entry. I assumed if it's not 2020 then the license does not apply but looking at other projects I think you're right..

Issue #314, opened 2018-03-28
SNS messaging: register HandlerMethodArgumentResolver with WebFluxConfigurer 


dgomezs:
We are in the process of upgrading our services to Spring boot 2.0 with the reactive stack and found a problem configuring SNS in that scenario. It seems not possible to register the resolvers as specified in the [documentation](https://cloud.spring.io/spring-cloud-aws/spring-cloud-aws.html#_annotation_driven_http_notification_endpoint). The only configuration provided extends **org.springframework.web.method.support.HandlerMethodArgumentResolver** rather than **org.springframework.web.reactive.result.method.HandlerMethodArgumentResolver** that you can override in WebFluxConfigurer.

How is SNS supposed to be configured with Webflux? Is there a workaround to make it work? We are using Finchley.M9 

Thank you very much for your help.

Isabek:
@maciejwalkowiak I would like to work on this issue. Could you assign it to me? Also, could you write a small description? Thanks.

maciejwalkowiak:
@Isabek the goal is to provide functionality from `org.springframework.cloud.aws.messaging.endpoint` to WebFlux in the same way it is currently provided for Spring WebMVC. The functionality is described at https://cloud.spring.io/spring-cloud-aws/reference/html/#annotation-driven-http-notification-endpoint

Isabek:
@maciejwalkowiak ok, please, assign it to me.

PR #611, opened 2020-06-20
Add WebFlux endpoint support for SNS.

Isabek:
Fixes gh-314

### DIFF
--- a/spring-cloud-aws-messaging/pom.xml
+++ b/spring-cloud-aws-messaging/pom.xml
@@ -48,6 +48,11 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework</groupId>
+			<artifactId>spring-webflux</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework</groupId>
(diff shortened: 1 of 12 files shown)


Isabek:
@maciejwalkowiak please, check this PR and give me feedback. I am not sure that my approach is correct.

maciejwalkowiak:
Thanks @Isabek for the PR. It took me a while to get into it - thanks for patience. There is quite a bit of duplication between regular handlers and reactive handlers - I think this could be refactored to reduce duplication.

There is one thing I missed before that I am afraid will prevent us from merging it before 3.0 - `AmazonSnsNotificationStatus` calls `confirmSubscription` on `AmazonSns` - this is a blocking call - which does not fit into reactive philosophy well. In 3.0 we will migrate to AWS SDK v2 which contains non-blocking client and this will let us create separate `ReactiveSnsNotificationStatus` with `confirmSubscription` method returning `Mono`.

maciejwalkowiak:
We will revisit WebFlux topic in 3.0. I am sorry but at this moment we are not able to merge this PR.

Issue #295, opened 2018-01-31
Enhancement: support native SQS messages in listener argument


MrBuddyCasino:
The @SqsListener annotation should support com.amazonaws.services.sqs.model.Message as method parameter. 

We want to store incoming events in DynamoDB including payload and headers, and the conversions that are done in QueueMessageUtils are making this considerably harder (String to MimeType / UUID etc.). Also we want to store events just like they come in, without any conversions or processing so there are less chances for bugs.

jarc0der:
Let me guys discover this one 
@maciejwalkowiak

maciejwalkowiak:
Go for it @jarc0der !

PR #612, opened 2020-06-21
Resolving native SQS message.

maciejwalkowiak:
Draft for now as I am not convinced if this is a right thing to do. Tests, docs etc are missing, but the implementation itself is working.

Fixes gh-295


spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/listener/QueueMessageHandler.java
```diff
@@ -88,6 +92,44 @@ public QueueMessageHandler() {
...
+
+		MessageHeaderAccessor headerAccessor = MessageHeaderAccessor
+				.getMutableAccessor(message);
+		headerAccessor.setHeader(
+				DestinationPatternsMessageCondition.LOOKUP_DESTINATION_HEADER,
+				lookupDestination);
+		headerAccessor.setLeaveMutable(true);
+
+		if (message instanceof SqsMessage) {
+			message = new SqsMessage((String) message.getPayload(),
```
maciejwalkowiak, line 114: This whole method had to be copied from parent class except this message creation. Parent class `handleMessage` ignores the type of original message and creates just a new `GenericMessage` - meaning we lose the object created in `QueueMessageUtils`.

voytech: Have You maybe considered something like this:
1. Do not modify QueueMessageHandler by overriding handleMessage. 
2. In QueueMessageUtils change method: 
`	public static Message<String> createMessage(
			com.amazonaws.services.sqs.model.Message message,
			Map<String, Object> additionalHeaders) {
		HashMap<String, Object> messageHeaders = new HashMap<>();
		messageHeaders.put(MESSAGE_ID_MESSAGE_ATTRIBUTE_NAME, message.getMessageId());
		messageHeaders.put(RECEIPT_HANDLE_MESSAGE_ATTRIBUTE_NAME,
				message.getReceiptHandle());

		messageHeaders.putAll(additionalHeaders);
		messageHeaders.putAll(getAttributesAsMessageHeaders(message));
		messageHeaders.putAll(getMessageAttributesAsMessageHeaders(message));

		return new GenericMessage<>(message.getBody(),
				new SqsMessageHeaders(messageHeaders));
	}`

So that it does not create GenericMessage with String payload, but rather with original message as a payload ? 
Resolving to original message or String can be then done by argument resolver.

In above approach You are not say 'impairing' intentions of AbstractMethodMessageHandler class which comes from  messaging, and You are changing code belonging only to spring-cloud-aws. 

I must say that this is just unconfirmed idea, I didn't check if changing QueueMessageUtils.createMessage to return GenericMessage<...aws.Message> is not breaking anything.....

voytech: But GenericMessage encapsulating entire original message... still workaround. Maybe not that good as I was thinking.

maciejwalkowiak: Thanks for an idea! I'll experiment with it and keep you updated

maciejwalkowiak:
@artembilan @markfisher could you please give us advice here? What would be the right approach to keep the original SQS message for further processing:

1. Extend `GenericMessage`:

```
public class SqsMessage extends GenericMessage<String> {

	private final Message originalMessage;

	public SqsMessage(String payload, MessageHeaders headers, Message originalMessage) {
		super(payload, headers);
		this.originalMessage = originalMessage;
	}

	public Message getOriginalMessage() {
		return originalMessage;
	}
}
```

2. Process `GenericMessage` with `com.amazonaws.services.sqs.model.Message` in `payload` field.

Both approaches work for what we are trying to achieve here - add ability to resolve original SQS message in `@SqsListener` annotated method parameters, but I am not sure if I am not missing something from bigger Spring Messaging picture.

artembilan:
Well, in other frameworks we go a different direction.
We have there some `MessageListener` abstraction which is specific to the broker.
In Spring Kafka it is like this: https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/listener/GenericMessageListener.java
Where `data` comes from here: https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/listener/MessageListener.java, which in your case could be that `com.amazonaws.services.sqs.model.Message`.

In Spring AMQP it is like this: https://github.com/spring-projects/spring-amqp/blob/master/spring-amqp/src/main/java/org/springframework/amqp/core/MessageListener.java
 
I do understand that logic here is made based on the `MessageHandler.handleMessage(Message<?>)` so it is too late to change it radically. And since you are tied with the Spring Messaging contract, I only can suggest to store that `com.amazonaws.services.sqs.model.Message` in the header. 
In Spring Integration we deal  with a header like this `IntegrationMessageHeaderAccessor.SOURCE_DATA` and populate it as an `org.springframework.amqp.core.Message` in the AMQP Inbound Channel Adapters and as a `ConsumerRecord` in Kafka Inbound Channel Adapters. 

So, when you work with this `@SqsListener` with an `com.amazonaws.services.sqs.model.Message` as one of the argument you could get its value from that header and be compatible with possible Spring Integration downstream.

Does it make sense?

P.S. Honestly I'm fully against any new `GenericMessage` extensions: it is hard to support all of them in other places, e.g. (de)serialiation.

maciejwalkowiak:
Thanks @artembilan! It's very useful. We will go with header for now and in 3.0 we will revisit it as we are going to revisit SQS integration internals anyway.

maciejwalkowiak:
Original message is now passed through `sourceData` header.

spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/listener/SqsMessageMethodArgumentResolver.java
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.messaging.Message;
+
+/**
+ * Resolves original SQS message object {@link com.amazonaws.services.sqs.model.Message}
+ * from Spring Messaging message object {@link Message}.
+ *
+ * @author Maciej Walkowiak
+ */
+public class SqsMessageMethodArgumentResolver implements
+		org.springframework.messaging.handler.invocation.HandlerMethodArgumentResolver {
```
eddumelendez, line 30: use import

spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/listener/SqsMessageMethodArgumentResolver.java
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright 2013-2019 the original author or authors.
```
eddumelendez, line 2: 2020

PR #613, opened 2020-06-21
Document handling exceptions thrown inside SqsListener

maciejwalkowiak:
Fixes gh-6


maciejwalkowiak:
Perhaps there's more to add but this is I believe better than nothing.

PR #617, opened 2020-06-29
Add support for STSAssumeRoleSessionCredentialsProvider

### LINKED ISSUE #136: Add support for STSAssumeRoleSessionCredentialsProvider
[assuming roles](http://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html) is another way to authenticate. 

`ContextCredentialsAutoConfiguration` could check for `cloud.aws.credentials.roleArn` and `cloud.aws.credentials.roleSessionName` and wrap the `AWSCredentialsProvider` created in `ContextConfigurationUtils.registerCredentialsProvider(..)` with a `STSAssumeRoleSessionCredentialsProvider`

anshlykov:
Closes #136

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-aws.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-aws.adoc
@@ -310,9 +310,10 @@ By default Spring Cloud AWS configures https://docs.aws.amazon.com/AWSJavaSDK/la
 
 If other credentials providers are configured, https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.html[DefaultAWSCredentialsProviderChain] is not used and Spring Cloud AWS configures following credentials chain:
 
-1. `AWSStaticCredentialsProvider` if `cloud.aws.credentials.access-key` is provided
-2. `EC2ContainerCredentialsProviderWrapper` if `cloud.aws.credentials.instance-profile` is set to `true`
-3. `ProfileCredentialsProvider` if `cloud.aws.credentials.profile-name` is provided
+1. `STSAssumeRoleSessionCredentialsProvider` if `cloud.aws.credentials.sts.role-arn` and `cloud.aws.credentials.sts.role-session-name` are provided
+2. `AWSStaticCredentialsProvider` if `cloud.aws.credentials.access-key` is provided
+3. `EC2ContainerCredentialsProviderWrapper` if `cloud.aws.credentials.instance-profile` is set to `true`
+4. `ProfileCredentialsProvider` if `cloud.aws.credentials.profile-name` is provided
 
 Spring Cloud AWS provides the following properties to configure the credentials setup for the whole application.
 
@@ -341,6 +342,18 @@ Spring Cloud AWS provides the following properties to configure the credentials 
 |cloud.aws.credentials.profile-path
 |`~/.aws/credentials`
 |The file path where the profile configuration file is located. Defaults to `~/.aws/credentials` if value is not provided
+
+|cloud.aws.credentials.sts.role-arn
+|arn:aws:iam::01234567890:role/sts-test
+|The arn of the role to be assumed
+
+|cloud.aws.credentials.sts.role-session-name
+|testRoleSessionName
+|An identifier for the assumed role session
+
+|cloud.aws.credentials.sts.role-session-duration-seconds
+|900
+|The duration for assume role sessions
 |===
 
 ===== Configuring region
--- a/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextCredentialsAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextCredentialsAutoConfiguration.java
@@ -25,6 +25,7 @@ import com.amazonaws.auth.AWSStaticCredentialsProvider;
 import com.amazonaws.auth.BasicAWSCredentials;
 import com.amazonaws.auth.DefaultAWSCredentialsProviderChain;
 import com.amazonaws.auth.EC2ContainerCredentialsProviderWrapper;
+import com.amazonaws.auth.STSAssumeRoleSessionCredentialsProvider;
 import com.amazonaws.auth.profile.ProfileCredentialsProvider;
 
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
@@ -67,6 +68,12 @@ public class ContextCredentialsAutoConfiguration {
 	private List<AWSCredentialsProvider> resolveCredentialsProviders(
 			AwsCredentialsProperties properties) {
 		List<AWSCredentialsProvider> providers = new ArrayList<>();
+
+		if (StringUtils.hasText(properties.getRoleArn())
+				&& StringUtils.hasText(properties.getRoleSessionName())) {
+			providers.add(new STSAssumeRoleSessionCredentialsProvider.Builder(
+					properties.getRoleArn(), properties.getRoleSessionName()).build());
+		}
 
 		if (StringUtils.hasText(properties.getAccessKey())
 				&& StringUtils.hasText(properties.getSecretKey())) {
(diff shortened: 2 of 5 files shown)


spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextCredentialsAutoConfiguration.java
```diff
@@ -85,6 +94,25 @@ public AWSCredentialsProvider awsCredentialsProvider(
...
+				stsClientBuilder.withRegion(regionProvider.getRegion().getName());
+			}
+
+			AWSCredentialsProvider provider = new STSAssumeRoleSessionCredentialsProvider.Builder(
+					properties.getRoleArn(), properties.getRoleSessionName())
+							.withRoleSessionDurationSeconds(
+									properties.getRoleSessionDurationSeconds())
+							.withStsClient(stsClientBuilder.build()).build();
+
+			providers.add(0, provider);
```
maciejwalkowiak, line 113: This provider configuration could be moved to the top of the method, then it would be easier to follow that it has a priority over other providers. Now I find it easy to miss that it adds it at index `0`.

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/properties/AwsCredentialsProperties.java
```diff
@@ -53,6 +54,21 @@
 	 */
 	private String profilePath;
 
+	/**
+	 * The arn of the role to be assumed.
+	 */
+	private String roleArn;
```
maciejwalkowiak, line 60: What do you think about moving all new properties to static inner class `StsCredentialsProperties` so that they are configured via properties: `cloud.aws.credentials.sts.roleArn` etc.?

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextCredentialsAutoConfiguration.java
anshlykov: Yes, I agree that it would be easier, but `STSAssumeRoleSessionCredentialsProvider` depends on the providers defined above. Besides, if temporary credentials are successfully created, other providers in the `AWSCredentialsProviderChain` will be ignored. It might make sense to overwrite other providers if authorization via STS is supposed. WDYT?

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/properties/AwsCredentialsProperties.java
anshlykov: Done

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextCredentialsAutoConfiguration.java
maciejwalkowiak: Thanks for explanation, I understand now.

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextCredentialsAutoConfiguration.java
```diff
@@ -85,6 +94,27 @@ public AWSCredentialsProvider awsCredentialsProvider(
 					: new ProfileCredentialsProvider(properties.getProfileName()));
 		}
 
+		AwsCredentialsProperties.StsCredentialsProperties stsProperties = properties
+				.getSts();
+		if (StringUtils.hasText(stsProperties.getRoleArn())
+				&& StringUtils.hasText(stsProperties.getRoleSessionName())) {
+
+			AWSSecurityTokenServiceClientBuilder stsClientBuilder = AWSSecurityTokenServiceClientBuilder
```
maciejwalkowiak, line 102: Since `sts` is an optional dependency, if user provides `roleArn` and `roleSessionName` but does not add dependency to `sts` code will fail with `NoClassDefFoundException`. 

We have to change it in a way that there is an attempt to use STS only if `sts` related classes are on the classpath.

docs/src/main/asciidoc/spring-cloud-aws.adoc
```diff
@@ -310,9 +310,10 @@ By default Spring Cloud AWS configures https://docs.aws.amazon.com/AWSJavaSDK/la
 
 If other credentials providers are configured, https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.html[DefaultAWSCredentialsProviderChain] is not used and Spring Cloud AWS configures following credentials chain:
 
-1. `AWSStaticCredentialsProvider` if `cloud.aws.credentials.access-key` is provided
-2. `EC2ContainerCredentialsProviderWrapper` if `cloud.aws.credentials.instance-profile` is set to `true`
-3. `ProfileCredentialsProvider` if `cloud.aws.credentials.profile-name` is provided
+1. `STSAssumeRoleSessionCredentialsProvider` if `cloud.aws.credentials.sts.role-arn` and `cloud.aws.credentials.sts.role-session-name` are provided
```
maciejwalkowiak, line 313: Big thanks for adding docs!

maciejwalkowiak:
@anshlykov thanks for your patience. We need to handle this use case when `sts` dependency is missing and then once @eddumelendez agrees I would be happy to merge it!

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/properties/AwsCredentialsProperties.java
```diff
@@ -93,4 +96,55 @@ public void setProfilePath(String profilePath) {
 		this.profilePath = profilePath;
 	}
 
+	public StsCredentialsProperties getSts() {
```
eddumelendez, line 99: Rename it to `Sts` and add javadoc please

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/ContextCredentialsAutoConfiguration.java
eddumelendez: logic for `sts` should be in another class with the proper `Conditional*` annotations and import it in this auto-configuration class just using the`AWSCredentialProvider`. In order to test it we can use `ClasspathExclusions` from `spring-boot` like [this example](https://github.com/spring-projects/spring-boot/blob/8e695b0cc5e834a01e9f12fa35af3b7fddddf58a/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/web/servlet/NoSpringWebFilterRegistrationBeanTests.java)

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/context/properties/AwsCredentialsProperties.java
```diff
@@ -93,4 +96,55 @@ public void setProfilePath(String profilePath) {
...
+	public void setSts(StsCredentialsProperties sts) {
+		this.sts = sts;
+	}
+
+	public static class StsCredentialsProperties {
+
+		/**
+		 * The arn of the role to be assumed.
+		 */
+		private String roleArn;
```
eddumelendez, line 112: @maciejwalkowiak WDYT about dropping `role` prefix for all the attributes in this class? Looks redundant to me

maciejwalkowiak:
Closed due to inactivity. While we believe this change is very welcome, credentials autoconfiguration have significantly changed since the time when PR was submitted. We will revisit this topic in 3.x

Issue #507, opened 2019-11-11
Set the default queueStopTimeout to >20 secs


madrian:
This will ensure all polls to queues even the long polling ones can be closed gracefully before the shutdown times out and prevents the `TimeoutException` to be encountered quite frequently like in Spring Boot.

https://stackoverflow.com/questions/58601625/prevent-spring-cloud-aws-messaging-from-trying-to-stop-the-queue

This doesn't mean either that all shutdowns will take >20 secs since the queues maybe short polling or already on their end of the poll when the shutdown is triggered.

Issue #504, opened 2019-10-29
Prevent spring-cloud-aws-messaging from trying to stop the queue


kkaushik90f:
I'm using spring-cloud-aws-messaging in a Spring Boot 2.1.9.RELEASE project.
I have SQS queue created manually in AWS.

It is being used like:

```
@SqsListener("${sqs.name.incoming}")
public void listen(String message) {
    ...
}
```
It works fine. But when I stop my application in IDE, or when the Spring Boot tests finish, it tries to stop the queue. It can't stop it and eventually times out. It throws this exception:

```
2019-10-29 15:40:07.949  WARN 10378 --- [       Thread-2] s.c.a.m.l.SimpleMessageListenerContainer : An exception occurred while stopping queue 'my-awesome-queue-name'

java.util.concurrent.TimeoutException: null
    at java.base/java.util.concurrent.FutureTask.get(FutureTask.java:204) ~[na:na]
    at org.springframework.cloud.aws.messaging.listener.SimpleMessageListenerContainer.waitForRunningQueuesToStop(SimpleMessageListenerContainer.java:161) ~[spring-cloud-aws-messaging-2.1.3.RELEASE.jar:2.1.3.RELEASE]
    at org.springframework.cloud.aws.messaging.listener.SimpleMessageListenerContainer.doStop(SimpleMessageListenerContainer.java:140) ~[spring-cloud-aws-messaging-2.1.3.RELEASE.jar:2.1.3.RELEASE]
    at org.springframework.cloud.aws.messaging.listener.AbstractMessageListenerContainer.stop(AbstractMessageListenerContainer.java:351) ~[spring-cloud-aws-messaging-2.1.3.RELEASE.jar:2.1.3.RELEASE]
    at org.springframework.cloud.aws.messaging.listener.SimpleMessageListenerContainer.stop(SimpleMessageListenerContainer.java:45) ~[spring-cloud-aws-messaging-2.1.3.RELEASE.jar:2.1.3.RELEASE]
    at org.springframework.cloud.aws.messaging.listener.AbstractMessageListenerContainer.stop(AbstractMessageListenerContainer.java:239) ~[spring-cloud-aws-messaging-2.1.3.RELEASE.jar:2.1.3.RELEASE]
[...]

PR #620, opened 2020-06-30
Fixes SQS Queues shutting down gracefully

mkatircioglu:
* Allow queueStopTimeout can be configurable via SimpleMessageListenerContainerFactory #gh-504
* Increase queueStopTimeout default value in milliseconds to be equal to waitTimeOut #gh-507

### DIFF
--- a/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/config/SimpleMessageListenerContainerFactory.java
+++ b/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/config/SimpleMessageListenerContainerFactory.java
@@ -29,6 +29,7 @@ import org.springframework.util.Assert;
 
 /**
  * @author Alain Sahli
+ * @author Mete Alpaslan Katırcıoğlu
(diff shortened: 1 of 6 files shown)


eddumelendez:
@mkatircioglu thanks for the PR! looks good to me but wondering if we can add a configuration property for `queueStopTimeout` which will offer more flexibility. WDYT?

mkatircioglu:
What kind of configuration property do you mean? Since there are some other properties in SimpleMessageListenerContainerFactory which might need to be configured too if we think to give some flexibility. @eddumelendez

mkatircioglu:
Hi @eddumelendez According to my comments, what is your latest opinion? Would you like me to update the PR?

maciejwalkowiak:
Since configuring custom `SimpleMessageListenerContainerFactory` is currently THE way to configure SQS listeners - we do not have configuration properties for SQS (yet) - this PR looks good to me. If you agree @eddumelendez I would be happy to merge it to 2.3.x branch.

eddumelendez:
Thanks for the input. Please go ahead! 👍🏽

PR #622, opened 2020-07-01
Improvements for paramstore and secretsmanager

eddumelendez:
- Move ParamStore autoconfig from starter to autoconfigure module
- Move SecretsManager autoconfig from starter to autoconfigure module
- Polish
- Add configuration metadata for paramstore and secretsmanager


maciejwalkowiak:
I believe placing SSM and ParameterStore configuration outside of `autoconfigure` package was not an accident. There is a valid use case to use one of these integrations without pulling any other extra dependency and I think that as long as `context`/`core` modules (which are not optional in `autoconfigure`) pull and set up S3, ec2, CloudFormation, SSM and ParameterStore configurations should stay in separate package. 

Maybe @jkuipers can chime in and share his motivation for putting it outside of `autoconfigure` module so that I don't guess ;-)

eddumelendez:
Thanks @maciejwalkowiak I do not much about the decisions for that but would be great to hear. As far as I can see there is room for improvement and indeed was wondering if should I send this PR or not.

jkuipers:
> I believe placing SSM and ParameterStore configuration outside of `autoconfigure` package was not an accident. There is a valid use case to use one of these integrations without pulling any other extra dependency and I think that as long as `context`/`core` modules (which are not optional in `autoconfigure`) pull and set up S3, ec2, CloudFormation, SSM and ParameterStore configurations should stay in separate package.
> 
> Maybe @jkuipers can chime in and share his motivation for putting it outside of `autoconfigure` module so that I don't guess ;-)

You guessed right. From https://github.com/spring-cloud/spring-cloud-aws/pull/308:
> I first added my BootstrapConfiguration to the existing autoconfiguration module, but that module has a hard dependency on the context module and configures things I don' personally care for when I just want the Parameter Store integration.
That would also prevent me from running an application locally while still retrieving its config from the Parameter Store.
I therefore added autoconfiguration for this new module to its starter instead.

eddumelendez:
> I first added my BootstrapConfiguration to the existing autoconfiguration module, but that module has a hard dependency on the context module and configures things I don' personally care for when I just want the Parameter Store integration.
That would also prevent me from running an application locally while still retrieving its config from the Parameter Store.
I therefore added autoconfiguration for this new module to its starter instead.

Agree, I think this can be simple but still working on that.

PR #623, opened 2020-07-01
Add user agent

eddumelendez:


spring-cloud-aws-core/src/main/java/org/springframework/cloud/aws/core/SpringCloudClientConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+import com.amazonaws.PredefinedClientConfigurations;
+
+/**
+ * @author Eddú Meléndez
+ */
+public final class SpringCloudClientConfiguration {
+
+	private static final String NAME = "spring-cloud-aws";
+
+	private static final String VERSION = "2.3.3";
```
maciejwalkowiak, line 29: I think you meant `2.2.3` here? Would be nice if we could populate this variable from Maven version.

eddumelendez: good catch! Agree, we can have that value from maven version. I will try to solve it and send another PR

maciejwalkowiak: In 2.2.3 I think it can stay as it is, perhaps another PR makes sense in 2.3 as there are quite a few changes already.

PR #631, opened 2020-07-17
Added support for custom type labels in headers from message attributes.

voytech:
Initial PR. 
- Added `CustomTypeMessageAttribute` class for handling type labels which may be present in SQS MessageAttribute type,
- Added handling of `Binary` and `String` MessageAttribute types when type label is present,
- Handling of `Number` MessageAttribute type stays as-is for **backward compatibility** (It does handle additional type label by assuming certain rules to be applied for label)


voytech:
Wrong target branch selected.

PR #632, opened 2020-07-17
Added support for custom type labels in headers from message attributes. Fixes #410

### LINKED ISSUE #410: Add support for SQS custom data types
Currently custom data types are silently ignored during transformation from MessageAttributes to MessageHeaders in QueueMessageUtils. Either custom data types should be supported fully or an Exception should be thrown.
If somebody uses SQS custom data type, which currently can only hold String or Binary values, it will get omitted in QueueMessageUtils and finding out why can be very time consumeing since it is omitted silently.

voytech:
Initial PR.

- Added CustomTypeMessageAttribute class for handling type labels which may be present in SQS MessageAttribute type,

- Added handling of Binary and String MessageAttribute types when type label is present,

- Handling of Number MessageAttribute type stays as-is for backward compatibility (It does handle additional type label by assuming certain rules to be applied for label)


voytech:
Decided to move changes from forked source branch 2.3.x to gh-410 to conform with conventions and house keeping guidelines

Issue #591, opened 2020-06-11
Deprecate @EnableX annotations


maciejwalkowiak:
Deprecate and eventually drop `@EnableX` annotations. Spring Cloud AWS has been created in pre-Spring Boot era where `@Enable` annotations was a very convenient way of configuring collection of beans in a declarative way. Since we are providing now auto-configuration for all our modules, we can drop `@Enable` annotations. It will bring following benefits:

- we can drop `Conditional` annotations provided by Spring Cloud AWS and use ones provided by Spring Boot
- we can drop a lot of factory beans code and use traditional Spring Boot autoconfigurations - this will make code much easier to approach for contributors
- we could use Spring Boot features when configuring integrations - since configuration will move completely to `autoconfigure` module (look SQS and SNS configuration)
- less confusion for users on how to use Spring Cloud AWS

I believe we should mark all `@Enable` annotations as deprecated in `2.3`, and drop them in `3.0`.

maciejwalkowiak:
@eddumelendez your feedback is appreciated.

eddumelendez:
As we were discussing, another benefit is `context` module can be optional too in 2.3 for `autoconfigure` module doing so. For further reference, new integrations should have their own module and/or autoconfiguration.

maciejwalkowiak:
Fixed in https://github.com/spring-cloud/spring-cloud-aws/pull/606, https://github.com/spring-cloud/spring-cloud-aws/pull/615, https://github.com/spring-cloud/spring-cloud-aws/pull/636

PR #636, opened 2020-07-24
Deprecate @EnableSqs and @EnableSns

maciejwalkowiak:
Part of #591.

- Deprecates `@EnableSQS` and `@EnableSns`.
- Exposes SQS configuration options through configuration properties - instead of overwriting factory beans:

```
cloud.aws.sqs:
  listener:
    max-number-of-messages: 5
    wait-timeout: 10
    visibility-timeout: 5
  handler:
    default-deletion-policy: ON_SUCCESS
```
... and more.
- Splits `MessagingAutoConfiguration` into `SnsAutoConfiguration` and `SqsAutoconfiguration` as it perfectly makes sense to use one of these two services independently.


spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/messaging/SnsAutoConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+		@Bean
+		public AmazonWebserviceClientFactoryBean<AmazonSNSClient> amazonSNS() {
+			return new AmazonWebserviceClientFactoryBean<>(AmazonSNSClient.class,
+					this.awsCredentialsProvider, this.regionProvider);
+		}
+
+	}
+
+	@Configuration(proxyBeanMethods = false)
+	@ConditionalOnClass(org.springframework.web.servlet.config.annotation.WebMvcConfigurer.class)
```
eddumelendez, line 76: avoid the fqn because it is already in the imports

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/messaging/SnsAutoConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+	static class SnsWebConfiguration {
+
+		@Bean
+		public WebMvcConfigurer snsWebMvcConfigurer(AmazonSNS amazonSns) {
+			return new WebMvcConfigurer() {
```
eddumelendez, line 81: replace by lambda expression

```
(resolvers) -> resolvers.add(getNotificationHandlerMethodArgumentResolver(amazonSns)
```

PR #637, opened 2020-07-25
Handling custom type labels in headers from Binary/String message attributes. Fix gh-410 [v2].

voytech:
This is another approach to solve gh-410.
First approach is here: https://github.com/spring-cloud/spring-cloud-aws/pull/633.
Original issue and discussion is here: https://github.com/spring-cloud/spring-cloud-aws/issues/410

### DIFF
--- a/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/core/QueueMessageUtils.java
+++ b/spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/core/QueueMessageUtils.java
@@ -112,8 +112,8 @@ public final class QueueMessageUtils {
 				messageHeaders.put(MessageHeaders.ID,
 						UUID.fromString(messageAttribute.getValue().getStringValue()));
 			}
-			else if (MessageAttributeDataTypes.STRING
-					.equals(messageAttribute.getValue().getDataType())) {
+			else if (messageAttribute.getValue().getDataType()
+					.startsWith(MessageAttributeDataTypes.STRING)) {
 				messageHeaders.put(messageAttribute.getKey(),
 						messageAttribute.getValue().getStringValue());
 			}
@@ -122,8 +122,8 @@ public final class QueueMessageUtils {
 				messageHeaders.put(messageAttribute.getKey(),
 						getNumberValue(messageAttribute.getValue()));
 			}
-			else if (MessageAttributeDataTypes.BINARY
-					.equals(messageAttribute.getValue().getDataType())) {
+			else if (messageAttribute.getValue().getDataType()
+					.startsWith(MessageAttributeDataTypes.BINARY)) {
 				messageHeaders.put(messageAttribute.getKey(),
 						messageAttribute.getValue().getBinaryValue());
 			}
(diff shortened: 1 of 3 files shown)


maciejwalkowiak:
I think this is the way to go. Considering we now give an option to just resolve original SQS message as a listener parameter, if someone is interested in more details about the specific message attribute it's still possible.

spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/core/QueueMessageUtils.java
```diff
@@ -112,8 +112,8 @@ else if (MessageHeaders.ID.equals(messageAttribute.getKey())) {
 				messageHeaders.put(MessageHeaders.ID,
 						UUID.fromString(messageAttribute.getValue().getStringValue()));
 			}
-			else if (MessageAttributeDataTypes.STRING
-					.equals(messageAttribute.getValue().getDataType())) {
+			else if (messageAttribute.getValue().getDataType()
+					.startsWith(MessageAttributeDataTypes.STRING)) {
```
eddumelendez, line 116: let's change it to `MessageAttributeDataTypes.STRING.startsWith(messageAttribute.getValue().getDataType())`

spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/core/QueueMessageUtils.java
```diff
@@ -122,8 +122,8 @@ else if (messageAttribute.getValue().getDataType()
 				messageHeaders.put(messageAttribute.getKey(),
 						getNumberValue(messageAttribute.getValue()));
 			}
-			else if (MessageAttributeDataTypes.BINARY
-					.equals(messageAttribute.getValue().getDataType())) {
+			else if (messageAttribute.getValue().getDataType()
+					.startsWith(MessageAttributeDataTypes.BINARY)) {
```
eddumelendez, line 126: let's change it to `MessageAttributeDataTypes. BINARY.startsWith(messageAttribute.getValue().getDataType())`

spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/core/QueueMessageUtils.java
voytech: getDataType may return „String” or string with label which is: „String.<type extensions labels>„

MessageAttributeDataTypes.STRING is always „String” 

If we replace as You suggest we will get:
„String”.startsWith(„String.<Here goes the label>”) which would be rejected, but we want to support everything which starts from „String”

spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/core/QueueMessageUtils.java
voytech: Here is the same as in comment above.

spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/core/QueueMessageUtils.java
eddumelendez: it was my fault, please ignore these comments. You are right.

maciejwalkowiak:
@voytech purely optional as you already put a lot of effort - could you also update docs?

voytech:
> @voytech purely optional as you already put a lot of effort - could you also update docs?

I have provided some info in docs. You can take a look if this is okey (or sth needs to be changed yet)

Issue #348, opened 2018-06-20
SimpleStorageResource not resolved in spring-boot 2.1.x


JurrianFahner:
In the documentation this snippet can be found:

```java
public class SimpleResourceLoadingBean {

    @Autowired
    private ResourceLoader resourceLoader;

    public void writeResource() throws IOException {
        Resource resource = this.resourceLoader.getResource("s3://myBucket/rootFile.log");
        WritableResource writableResource = (WritableResource) resource;
        try (OutputStream outputStream = writableResource.getOutputStream()) {
            outputStream.write("test".getBytes());
        }
    }
}
```

This works in Finchley.RC2 as expected. But when I run the same code with only spring cloud updated to Finchley.RELEASE, I get the following class cast exception: `org.springframework.web.context.support.ServletContextResource cannot be cast to org.springframework.core.io.WritableResource`

In my debugger I see that the Resource should be an instance of a SimpleStorageResource, instead it is an instance of a ServletContextResource. 

That is caused by the resourceLoader which shoud be a org.springframework.cloud.aws.core.io.s3.PathMatchingSimpleStorageResourcePatternResolver, instead it is an instance of a AnnotationConfigServletWebServerApplicationContext.

I use spring-boot 2.0.3

aemruli:
Thanks  @JurrianFahner for reporting. This is unfortunately a regression as we have utilized some new Spring features for loading resources (to solve some other bugs). It seems that this does not work with the Web Application context. I will add test cases with a web application context and see how to solve it. 

In the meantime you could inject a org.springframework.cloud.aws.core.io.s3.SimpleStorageProtocolResolver and use that directly (pass in the location and the application context to the resolve method)

JurrianFahner:
Well that's not so easy. Because SimpleStorageProtocolResolver needs an S3 client with credentials.

So I use RC2, until this is fixed. My project is in dev phase, so I have some time. I imagine that it will be fixed soon.

sebastianreloaded:
This makes the S3 resource loader indeed unusable. RC2 does work though.

Is there a plan to fix this with 2.0.x ?

sebastianreloaded:
Is this related to #384 ?

JurrianFahner:
> Is this related to #384 ?

Nope, because I didn't use devtools. So removing devtools wouldn't fix the issue.

denis111:
I have used this inline workaround for now

```java
Resource resource = resourceLoader.getResource(location);

[...]

PR #638, opened 2020-07-25
Fix SimpleStorageProtocolResolver bean registration.

maciejwalkowiak:
Fixes gh-348


maciejwalkowiak:
I did not manage to reproduce this bug in integration tests. Complete Spring Boot based application that reproduces this bug can be found here: https://github.com/PP-etc/spring-boot-with-s3-example/blob/master/src/main/java/com/pavelpolyakov/s3example/S3ExampleApplication.java#L33

PR #659, opened 2020-09-14
Change the default charset to  UTF-8 to support Special Characters

chsraju:
Charset.forName(WebUtils.DEFAULT_CHARACTER_ENCODING) to
Charset.defaultCharset() to support special characters using UTF-8


chsraju:
resubmitting the push request

Issue #645, opened 2020-08-21
@NotificationMessageMapping handler cannot deserialize Chinese characters


jpthompson23:
**Describe the bug**
I publish a notification to SNS that contains a JSON object with Chinese characters in some of the fields, such as:
`{ "documentTitle": "口廿竹十火 - new" }`

And the NotificationMessageMapping handler deserializes this as:
`{ "documentTitle": "????? - new" }`

**Sample**
```
    // try this with an SNS message that contains Chinese characters:
    @NotificationMessageMapping
    public void handleNotification(@NotificationMessage JsonNode message, @NotificationSubject String subject) {
        System.out.println(message);
    }
```

PR #660, opened 2020-09-15
Fix the issue with special characters - Changed the default charset to UTF-8 to support special characters

chsraju:
Changed the default charset to UTF-8 to support special characters


chsraju:
This is to address the below reported issue
https://github.com/spring-cloud/spring-cloud-aws/issues/645#issue-683832969

maciejwalkowiak:
Thanks @chsraju for contribution! Is it possible to also add a test that verifies this change?

spring-cloud-aws-messaging/src/main/java/org/springframework/cloud/aws/messaging/endpoint/NotificationMessageHandlerMethodArgumentResolver.java
```diff
@@ -127,7 +126,7 @@ public InputStream getBody() throws IOException {
 
 		private Charset getCharset() {
 			return this.mediaType.getCharset() != null ? this.mediaType.getCharset()
-					: Charset.forName(WebUtils.DEFAULT_CHARACTER_ENCODING);
+					: Charset.defaultCharset();
```
maciejwalkowiak, line 129: This will use default charset from OS. According to SNS docs, strings are encoded with UTF-8 so perhaps instead we should use here `StandardCharsets.UTF_8`?

maciejwalkowiak (changes requested):

Issue #664, opened 2020-09-16
Add a new ses module


eddumelendez:
**Is your feature request related to a problem? Please describe.**
Simple Email Service is already provided as an integration in the project but currently depends on `context` module. 

**Describe the solution you'd like**
 Instead, we can provide a `ses` module and starter. Since it is targeted to 2.3.x should be backward compatible with existing configuration.

**Describe alternatives you've considered**

**Additional context**

PR #669, opened 2020-09-16
Add ses module

eddumelendez:
Closes gh-664

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/mail/SesAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/mail/SesAutoConfiguration.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2013-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.aws.autoconfigure.mail;
+
+import javax.mail.Session;
+
+import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.services.simpleemail.AmazonSimpleEmailService;
+import com.amazonaws.services.simpleemail.AmazonSimpleEmailServiceClient;
+
+import org.springframework.beans.factory.ObjectProvider;
+import org.springframework.boot.autoconfigure.AutoConfigureAfter;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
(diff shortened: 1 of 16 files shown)


spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/mail/SesAutoConfiguration.java
```diff
@@ -0,0 +1,99 @@
...
+ */
+@Configuration(proxyBeanMethods = false)
+@AutoConfigureAfter(MailSenderAutoConfiguration.class)
+@ConditionalOnClass({ AmazonSimpleEmailService.class, MailSender.class })
+@ConditionalOnMissingBean(MailSender.class)
+@Import(ContextCredentialsAutoConfiguration.class)
+@EnableConfigurationProperties(SesProperties.class)
+@ConditionalOnProperty(name = "spring.cloud.aws.ses.enabled", havingValue = "true",
+		matchIfMissing = true)
+public class SesAutoConfiguration {
```
maciejwalkowiak, line 62: What happens if both old mail auto-config and new SES auto-config are enabled? I cannot figure it out just by looking at the code. Is there an order somewhere baked in?

maciejwalkowiak (changes requested):
Please include also changes in docs. It's difficult to catch up with docs at the later stage.

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/mail/SesAutoConfiguration.java
eddumelendez: good catch! I have added `@ConditionalOnMissingClass("org.springframework.cloud.aws.mail.simplemail.SimpleEmailServiceJavaMailSender")`

maciejwalkowiak: Assuming that `context` at least now is always included in user's projects, this would mean that `SesAutoConfiguration` never gets triggered, right?

Maybe we can make it in a way, that `SesAutoConfiguration` has `@AutoConfigureBefore(SimpleEmailAutoConfiguration.class)`? Then in case user intentionally adds SES Starter, old `SimpleEmailAutoConfiguration` will not fire because the `MailSender` bean will be already defined. What do you think?

eddumelendez: > Assuming that `context` at least now is always included in user's projects, this would mean that `SesAutoConfiguration` never gets triggered, right?

That's right. At least during 2.3

I like the idea. Will do the changes later today 👍🏽

eddumelendez: changes were applied

maciejwalkowiak (approved):
Great work @eddumelendez! When you merge it, please only write longer commit message explaining what and why are we doing here.

maciejwalkowiak:
I think it would be great if we add documentation together with PRs with code changes.

eddumelendez:
@maciejwalkowiak I have rebase the branch and update the docs. Could you review it again, please?

maciejwalkowiak (approved):
Apologies for long delay.

buckett:
Thanks for this, I've just had to sort out SES (in Spring Cloud AWS) and it's been a little tricky, hopefully this will make it allot easier in future for everyone.

Issue #123, opened 2016-01-13
Feature Request: Spring Security integration with Amazon Cognito


matsev:
[Amazon Cognito](https://aws.amazon.com/cognito/):

> Amazon Cognito is a service that makes it easy to save mobile user data, such as app preferences or game state, in the AWS Cloud without writing any backend code or managing any infrastructure. Cognito offers mobile identity management and data synchronization across devices. You can save data locally on users’ devices allowing your applications to work even when the devices are offline. You can also synchronize data across a user’s devices so that their app experience will be consistent regardless of the device they use. With Amazon Cognito, you can focus on creating great app experiences instead of having to worry about building and managing a backend solution to handle user authentication, network state, storage, and sync.

As can be seen, Amazon focuses on not having any backend implementation (typically, your mobile application calls other Amazon services directly after using Cognito), but it would be beneficial to implement your own backend service and yet leveraging on Amazon Cognito for user authentication. Like Spring Social, Cognito offers authentication integration with Facebook, Twitter, OpenId and others, but [Amazon Cognito Identity](http://docs.aws.amazon.com/cognito/devguide/identity/) also offers anonymous authentication (that later can be promoted to a real authentication for the same user). In contrast to Spring Social, Cognito Identoty also enables [Amazon Cognito Sync](http://docs.aws.amazon.com/cognito/devguide/sync/) service that among other things synchronize the state of your mobile app with the AWS cloud. 

In the blog post [Use Amazon Cognito in your website for simple AWS authentication](http://mobile.awsblog.com/post/TxBVEDL5Z8JKAC/Use-Amazon-Cognito-in-your-website-for-simple-AWS-authentication) the authentication flow is outlined for some programming languages, including Java.

matsev:
See also the [AWSCognitoSampleDeveloperAuthenticationSample](https://github.com/awslabs/amazon-cognito-developer-authentication-sample) sample project that is provided by the [Amazon Web Services - Labs](https://github.com/awslabs) GitHub account.

suchitgupta:
This sounds great. Sorry to bother you when is this expected to be completed

andrefaria:
This might be helpful when implementing http://blog.kdgregory.com/2016/12/server-side-authentication-with-amazon.html

andrefaria:
Another open source that might be helpful https://github.com/IxorTalk/ixortalk.aws.cognito.jwt.security.filter

DmytroRybka:
[...]

PR #680, opened 2020-09-20
Add cognito integration

eddumelendez:
See gh-123

### DIFF
--- a/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/security/CognitoAuthenticationAutoConfiguration.java
+++ b/spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/security/CognitoAuthenticationAutoConfiguration.java
@@ -42,7 +42,7 @@ import org.springframework.security.oauth2.jwt.NimbusJwtDecoder;
  * @author Eddú Meléndez
  * @since 2.3
  */
-@Configuration
+@Configuration(proxyBeanMethods = false)
 @AutoConfigureBefore(OAuth2ResourceServerAutoConfiguration.class)
 @EnableConfigurationProperties(CognitoAuthenticationProperties.class)
 @ConditionalOnProperty(prefix = "spring.cloud.aws.security.cognito", name = "enabled",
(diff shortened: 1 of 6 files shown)


eddumelendez:
Doc related https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-tokens-with-identity-providers.html#amazon-cognito-user-pools-using-the-id-token

spring-cloud-aws-autoconfigure/src/main/java/org/springframework/cloud/aws/autoconfigure/security/CognitoAuthenticationAutoConfiguration.java
```diff
@@ -0,0 +1,103 @@
...
+		return jwtDecoder;
+	}
+
+	static class OnUserPoolIdAndRegionPropertiesCondition extends AllNestedConditions {
+
+		OnUserPoolIdAndRegionPropertiesCondition() {
+			super(ConfigurationPhase.PARSE_CONFIGURATION);
+		}
+
+		@ConditionalOnProperty("spring.cloud.aws.security.cognito.userPoolId")
```
maciejwalkowiak, line 91: Properties should not be referenced with camel case. Here: `spring.cloud.aws.security.cognito.user-pool-id`

maciejwalkowiak (commented):
If I got it right, with this integration we are saving user from setting Spring Security OAuth config like explained here: https://www.baeldung.com/spring-security-oauth-cognito - is this correct?

I am wondering if we could do more - for example map Cognito groups to GrantedAuthorities (I am not even sure if this is a good idea)

eddumelendez:
> If I got it right, with this integration we are saving user from setting Spring Security OAuth config like explained here: https://www.baeldung.com/spring-security-oauth-cognito - is this correct?

Yes, that's right

> I am wondering if we could do more - for example map Cognito groups to GrantedAuthorities (I am not even sure if this is a good idea)

Not sure either. We can receive feedback once integrated

eddumelendez:
I have tested it and it is working after some polish

maciejwalkowiak (approved):

PR #681, opened 2020-09-21
Add elasticache redis

eddumelendez:


eddumelendez:
doc related 
[1] https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/Clusters.ViewDetails.html
[2] https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/Endpoints.html#Endpoints.Find.Redis
[3] https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/Endpoints.html#Endpoints.Find.RedisCluster

eddumelendez:
I have changed the logic to resolve if it is a cluster or not due to ClusterCache.getConfiguration is not returned. Also, I have created a ticket https://github.com/aws/aws-sdk-java/issues/2419

eddumelendez:
I am thinking about having a property `cluster-enabled` @maciejwalkowiak WDYT?

mkatircioglu:
Wouldn't it better to name module as spring-cloud-starter-aws-elasticache?

eddumelendez:
Currently, we will be building on top of `spring-boot-starter-data-redis` in order to provide the right configuration for elasticache redis nodes and clusters. Having a `spring-cloud-starter-aws-elasticache` module will make sense if we want to drop `memcache` support which is something we have not figure it out yet and having `redis` as the default which sometimes can be confused. I prefer to be explicit here.

/cc @maciejwalkowiak

PR #688, opened 2020-10-03
Missing enabled tests

MatejNedic:
Param store and Secret store autoconfiguration have missing enabled tests but since migrating to contextRunner is in another ticket and currently in 2.3.x it wasn't included.


maciejwalkowiak (approved):
Thanks @MatejNedic for another contribution!

Issue #655, opened 2020-09-13
Add support for import configuration via spring.config.import for Secrets Manager


eddumelendez:
**Is your feature request related to a problem? Please describe.**
Read Supporting Additional Locations https://spring.io/blog/2020/08/14/config-file-processing-in-spring-boot-2-4

**Describe the solution you'd like**
Use `aws-secretsmanager:` as a prefix

**Describe alternatives you've considered**

**Additional context**

Issue #515, opened 2019-12-12
Allow adding any arbitrary AWS Secrets Manager secrets 


tvrmsmith:
It would be great if we could add additional AWS Secrets Manager secrets as property sources. Currently, I'd like to use the RDS secrets feature as well as other secrets, but the current way of adding secrets makes it difficult to do this.

It could be configured something like:
```
aws:
  secretsmanager:
    additional-secrets:
      - /secret/secret1
      - /secret/secret2
```

maciejwalkowiak:
@tvrmsmith just so that I understand it correctly - these additional secrets would contains raw values and everything after prefix (`/secret/`) would he a property name?

eddumelendez:
I think [this link](https://aws.amazon.com/blogs/security/rotate-amazon-rds-database-credentials-automatically-with-aws-secrets-manager/) can be useful to understand the proposal. @tvrmsmith is what you meant?

maciejwalkowiak:
In case of RDS password I would like us to use this library in 3.0 https://github.com/aws/aws-secretsmanager-jdbc. I have it working in the prototype (... and made video about it how to use this library ;-) https://www.youtube.com/watch?v=gIbr6-AR6T8)

nhomble:
Instead of an array of strings, maybe we should be a little more flexible and allow users to configure the desired name as well? I think we're going to hit a wall if we only depend on the library to interpret the secret name from the path. Maybe like

```yml
aws:
  secretsmanager:
    additional-secrets:
      secret1:
        path: /secret/secret1
      different-name:
        path: /secret/secret2
```

nickcaballero:
Agreed with @nhomble. I have a similar implementation of this via a `PropertyResolver`.

PR #721, opened 2020-11-13
Add aws-secretsmanager: prefix config import

eddumelendez:
In `spring-boot` 2.4, `Volume Mounted Config Directory Trees` was
added. This commit introduces the prefix `aws-secretsmanager:` which
will resolve the values given the configuration properties supported
by secrets manager integration. Also, if keys are added after the
prefix then just these will be resolved.

Use: `aws-secretsmanager:` or `aws-secretsmanager:my-secret-key` or
`aws-secretsmanager:my-secret-key;my-anoter-secret-key`

Closes gh-655
Closes gh-515

### DIFF
--- a/docs/src/main/asciidoc/secrets-manager.adoc
+++ b/docs/src/main/asciidoc/secrets-manager.adoc
@@ -69,3 +69,19 @@ dots, dashes, forward slashes, backward slashes and underscores next to alphanum
 |`true`
 |Can be used to disable the Secrets Manager Configuration support even though the auto-configuration is on the classpath.
 |===
+
+In `spring-cloud` `2020.0.0` (aka Ilford), the bootstrap phase is no longer enabled by default. In order
+enable it you need an additional dependency:
+
+[source,xml,indent=0]
+----
+<dependency>
+  <groupId>org.springframework.cloud</groupId>
+  <artifactId>spring-cloud-starter-bootstrap</artifactId>
+  <version>{spring-cloud-version}</version>
+</dependency>
+----
+
+However, starting at `spring-cloud-aws` `2.3`, allows import default aws' secretsmanager keys
+(`spring.config.import=aws-secretsmanager:`) or individual keys
+(`spring.config.import=aws-secretsmanager:secret-key;other-secret-key`)
--- a/spring-cloud-aws-secrets-manager-config/src/main/java/org/springframework/cloud/aws/secretsmanager/AwsSecretsManagerPropertySourceLocator.java
+++ b/spring-cloud-aws-secrets-manager-config/src/main/java/org/springframework/cloud/aws/secretsmanager/AwsSecretsManagerPropertySourceLocator.java
@@ -31,7 +31,6 @@ import org.springframework.core.env.CompositePropertySource;
 import org.springframework.core.env.ConfigurableEnvironment;
 import org.springframework.core.env.Environment;
 import org.springframework.core.env.PropertySource;
-import org.springframework.util.ReflectionUtils;
 
 /**
  * Builds a {@link CompositePropertySource} with various
@@ -41,17 +40,18 @@ import org.springframework.util.ReflectionUtils;
  *
  * @author Fabio Maia
  * @author Matej Nedic
+ * @author Eddú Meléndez
  * @since 2.0.0
  */
 public class AwsSecretsManagerPropertySourceLocator implements PropertySourceLocator {
 
-	private String propertySourceName;
+	private final String propertySourceName;
 
-	private AWSSecretsManager smClient;
+	private final AWSSecretsManager smClient;
 
-	private AwsSecretsManagerProperties properties;
+	private final AwsSecretsManagerProperties properties;
 
-	private final Set<String> contexts = new LinkedHashSet();
+	private final Set<String> contexts = new LinkedHashSet<>();
 
 	private Log logger = LogFactory.getLog(getClass());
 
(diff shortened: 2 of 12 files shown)


docs/src/main/asciidoc/secrets-manager.adoc
```diff
@@ -69,3 +69,19 @@ dots, dashes, forward slashes, backward slashes and underscores next to alphanum
...
+<dependency>
+  <groupId>org.springframework.cloud</groupId>
+  <artifactId>spring-cloud-starter-bootstrap</artifactId>
+  <version>{spring-cloud-version}</version>
+</dependency>
+----
+
+However, starting at `spring-cloud-aws` `2.3`, allows import default aws' secretsmanager keys
+(`spring.config.import=aws-secretsmanager:`) or individual keys
+(`spring.config.import=aws-secretsmanager:secret-key;other-secret-key`)
```
maciejwalkowiak, line 87: It would be nice to document that it can be used with `optional` prefix.

spring-cloud-starter-aws-secrets-manager-config/src/main/java/org/springframework/cloud/aws/autoconfigure/secretsmanager/AwsSecretsManagerConfigDataLoader.java
```diff
@@ -0,0 +1,49 @@
...
+	@Override
+	public ConfigData load(ConfigDataLoaderContext context, AwsSecretsManagerConfigDataResource resource) {
+		try {
+			AWSSecretsManager ssm = context.getBootstrapContext().get(AWSSecretsManager.class);
+			AwsSecretsManagerPropertySource propertySource = new AwsSecretsManagerPropertySource(resource.getContext(),
+					ssm);
+			propertySource.init();
+			return new ConfigData(Collections.singletonList(propertySource));
+		}
+		catch (Exception e) {
```
maciejwalkowiak, line 44: We should respect here that config data can be marked as optional for example: `spring.config.import=optional:aws-secretsmanager:`.

Could we handle it in such a way?
```java
catch (Exception e) {
	if (!resource.isOptional()) {
		throw new ConfigDataResourceNotFoundException(resource, e);
	}
	if (logger.isWarnEnabled()) {
		logger.warn("Unable to load AWS secret from " + resource.getContext() + ". " + e.getMessage());
	} else {
		logger.trace("Unable to load AWS secret from " + resource.getContext(), e);
	}
	return null;
}
```
We would have to change `AwsSecretsManagerPropertySource` to not catch and log `ResourceNotFoundException`.

spring-cloud-starter-aws-secrets-manager-config/src/main/java/org/springframework/cloud/aws/autoconfigure/secretsmanager/AwsSecretsManagerConfigDataLoader.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.boot.context.config.ConfigDataLoader;
+import org.springframework.boot.context.config.ConfigDataLoaderContext;
+import org.springframework.boot.context.config.ConfigDataResourceNotFoundException;
+import org.springframework.cloud.aws.secretsmanager.AwsSecretsManagerPropertySource;
+
+/**
+ * @author Eddú Meléndez
+ * @since 2.3.0
+ */
+public class AwsSecretsManagerConfigDataLoader implements ConfigDataLoader<AwsSecretsManagerConfigDataResource> {
```
maciejwalkowiak, line 33: The issue is right now that it silently fails when referenced secrets do not exist. Check next comment to see how I think we can tackle it.

spring-cloud-starter-aws-secrets-manager-config/src/main/java/org/springframework/cloud/aws/autoconfigure/secretsmanager/AwsSecretsManagerConfigDataLocationResolver.java
```diff
@@ -0,0 +1,148 @@
...
+
+		AwsSecretsManagerProperties properties = loadConfigProperties(resolverContext.getBinder());
+
+		AwsSecretsManagerPropertySources sources = new AwsSecretsManagerPropertySources(properties, log);
+
+		List<String> contexts = location.getValue().equals(PREFIX)
+				? sources.getAutomaticContexts(profiles.getAccepted())
+				: getCustomContexts(location.getNonPrefixedValue(PREFIX));
+
+		ArrayList<AwsSecretsManagerConfigDataResource> locations = new ArrayList<>();
```
maciejwalkowiak, line 87:
```suggestion
		List<AwsSecretsManagerConfigDataResource> locations = new ArrayList<>();
```

spring-cloud-starter-aws-secrets-manager-config/pom.xml
```diff
@@ -44,6 +44,10 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-aws-secrets-manager-config</artifactId>
 		</dependency>
+		<dependency>
```
maciejwalkowiak, line 47: Why do we need a web starter?

maciejwalkowiak (changes requested):
Great work @eddumelendez! I left few comments - we need to figure out the proper handling of `optional` config data import.

Btw, Vault integration could be helpful reference for us in general for this feature: https://github.com/spring-cloud/spring-cloud-vault/commit/451ccb0bb3ba991d8a06bc4e5f7efce0e8ae359c

It would be great if you could add Javadocs but they can be also added later. Once this one is merged I will create a PR with a sample that I did to verify how/if it works.

spring-cloud-starter-aws-secrets-manager-config/src/main/java/org/springframework/cloud/aws/autoconfigure/secretsmanager/AwsSecretsManagerConfigDataLoader.java
eddumelendez: good point! I miss that part

spring-cloud-aws-secrets-manager-config/src/main/java/org/springframework/cloud/aws/secretsmanager/AwsSecretsManagerPropertySourceLocator.java
```diff
@@ -78,56 +78,20 @@ public AwsSecretsManagerPropertySourceLocator(AWSSecretsManager smClient, AwsSec
 
 		ConfigurableEnvironment env = (ConfigurableEnvironment) environment;
 
-		String appName = properties.getName();
-
-		if (appName == null) {
-			appName = env.getProperty("spring.application.name");
-		}
+		AwsSecretsManagerPropertySources sources = new AwsSecretsManagerPropertySources(properties, logger);
```
maciejwalkowiak, line 81: As far as I can see you have ignored checking `spring.application.name` which potentially will break users using old way of using Secrets Manager.

maciejwalkowiak:
`failFast` and `optional` did not work well for me. I fixed it and polished but I can't create a PR to your branch cause you pushed it into your fork, so please take a look at the commit: https://github.com/spring-cloud/spring-cloud-aws/commit/21011e629c1e176894c707368d9dc9c871743e02 and maybe cherry pick it

maciejwalkowiak (approved):

PR #730, opened 2020-11-25
Disable ContextInstanceDataAutoConfiguration by default.

maciejwalkowiak:
## :loudspeaker: Type of change

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

This change makes `ContextInstanceDataAutoConfiguration` an opt-in feature - to use it users have to explicitly set the `cloud.aws.instance.data.enabled` to `true`.

It is a breaking change.

## :bulb: Motivation and Context

ContextInstanceDataAutoConfiguration has a condition to run only in AWS environment. This condition is expensive to run outside of AWS environment and it slows down application startup several seconds.

## :green_heart: How did you test it?

Integration tests.

## :pencil: Checklist

- [x] I reviewed submitted code
- [x] I added tests to verify changes
- [x] I updated reference documentation to reflect the change
- [x] All tests passing
- [ ] No breaking changes


eddumelendez (approved):
🚀

